PR #33, opened 2014-12-19
Workaround for the VS10 min macro problem

ptoharia:
VS10 has a min macro that gives a building error. Placing parenthesis around std::min gives a clean workaround.


eile:
@ptoharia : #34 readded the NOMINMAX. Please test and reopen a PR if it still does not work.

ptoharia:
@eile ok, sorry for not testing it but I've been on vacation until today

Issue #29, opened 2014-12-11
Remote connection broker


eile:
# Use Case

A remote publisher needs a subscriber to subscribe to him directly, since ZeroConf is not available between the local and remote nodes. 

# Design

Two new classes in a nested zeq::connection namespace provide a Broker on the subscriber side and a Service on the publisher side. They are implemented internally using the request (service) - reply (broker) pattern. The broker has a Subscriber, to which it adds connections it receives from the service. The Service has a subscribe method taking a Publisher (address).
![UML Diagram](https://raw.githubusercontent.com/HBPVIS/zeq/master/doc/broker.png)

# Implementation

* [x] Broker/Service classes
* [x] #30: New Receiver base class implementing the zmq_poll logic. Base class of subscriber and broker. Has a Receiver( Receiver& shared ) ctor to allow one receive() poll all receiver sockets.

eile:
Implemented.

PR #34, opened 2015-01-05
connection::Broker and Service implementation

eile:
See #29 for background


eile:
@bhentschel @hernando @tribal-tec @bilgili @cnowke @haenel Please review.

tests/connection/broker.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <lunchbox/sleep.h>
+#include <lunchbox/thread.h>
+#include <lunchbox/servus.h>
+#include <boost/bind.hpp>
+
+using boost::lexical_cast;
+
+const unsigned short port = (lunchbox::RNG().get<uint16_t>() % 60000) + 1024;
+bool received = false;
```
bhentschel: a global variable, really?
Might want to put this into the Service class w/ dedicated getter and make it atomic since it's accessed by two threads concurrently.

tests/connection/broker.cpp
```diff
@@ -0,0 +1,68 @@
...
+    service.start();
+
+    // Using a different scheme so zeroconf resolution does not work
+    zeq::Publisher publisher( lunchbox::URI( "bar://*:" +
+                                          lexical_cast< std::string >( port )));
+    const std::string address =
+        std::string( "127.0.0.1:" ) + lexical_cast< std::string >( port+1 );
+    BOOST_CHECK( zeq::connection::Service::subscribe( address, publisher ));
+
+    for( size_t i = 0; i < 10 && !received; ++i )
```
bhentschel: 10 as a value seems a bit arbitrary. Hand the threshold some self-describing name and use it consistently here and in the receiving thread? Going for a formulation while(!service->TimedOut()) with bool Service::TimedOut() which combines the (service's) iteration count and the received flag might be an even better option.

tests/connection/broker.cpp
eile: It's just for the unit test.

zeq/connection/service.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <zeq/publisher.h>
+#include <lunchbox/log.h>
+#include <lunchbox/sleep.h>
+#include <zmq.h>
+
+namespace zeq
+{
+namespace connection
+{
+bool Service::subscribe( const std::string& broker, const Publisher& publisher )
```
bhentschel: broker ==> brokerAddress

tests/connection/broker.cpp
bhentschel: Ack!
Yet this was targeted at general test readability, not so much a general "global-variable-rant" ;)

zeq/connection/broker.h
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * For simplicity, only a single Receiver is handled by a Broker. The
+     * implementation should be extended if multiple receivers shall be
+     * handled.
+     *
+     * @param address the zmq reply socket address to be used.
+     * @param receiver the Receiver to manage.
+     * @throw std::runtime_error when the zmq setup failed.
+     */
+    ZEQ_API Broker( const std::string& address, Receiver& receiver );
```
bhentschel: If I get this correctly, this enforces a 1:1 relationship between broker and receiver. Hence, it's not possible to have a single broker cater for multiple receivers, right? I wonder if that would be desirable.

zeq/connection/service.cpp
eile: done.

tests/connection/broker.cpp
hernando: I'm also in for making received a member of Service for readability.
I'm not an expert in low-level details on memory models but I think that making it atomic is not necessary as lunchbox::sleep should imply a sufficient memory barrier.

zeq/connection/broker.h
eile: Right. It can be added later, but in the spirit of KISS (there is no use case for 1:n right now) we decided to have a simpler implementation.

tests/connection/broker.cpp
eile: done. Also moved the final test after the thread join, which definitely will have the proper barrier.

zeq/connection/broker.h
bhentschel: Sorry, my fault. In my mindset it's still (1 socket <=> 1 message type/semantic message channel), which entails a number of receivers per component, which in turn means handling a number of brokers. Hard to get rid of old habits, even over the turn of the year...

zeq/connection/service.cpp
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    const std::string& address = publisher.getAddress();
+    zmq_msg_t request;
+    zmq_msg_init_size( &request, address.size( ));
+    memcpy( zmq_msg_data( &request ), address.c_str(), address.size( ));
+
+    if( zmq_msg_send( &request, socket, 0 ) == -1 )
+    {
+        LBINFO << "Can't send connection request " << address << " to "
```
tribal-tec: IIRC, zmq_msg_close is needed always after zmq_msg_init*

zeq/connection/service.cpp
```diff
@@ -0,0 +1,65 @@
...
+               << std::endl;
+        return false;
+    }
+    zmq_msg_close( &request );
+
+    zmq_msg_t reply;
+    zmq_msg_init( &reply );
+    if( zmq_msg_recv( &reply, socket, 0 )  == -1 )
+    {
+        LBINFO << "Can't receive connection reply from " << brokerAddress
```
tribal-tec: dito

zeq/publisher.h
```diff
@@ -1,7 +1,7 @@
 
-/* Copyright (c) 2014, Human Brain Project
- *                     Daniel Nachbaur <daniel.nachbaur@epfl.ch>
- *                     Stefan.Eilemann@epfl.ch
+/* Copyright (c) 20142015, Human Brain Project
```
tribal-tec: that's a big year!

tests/connection/broker.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <lunchbox/sleep.h>
+#include <lunchbox/thread.h>
+#include <lunchbox/servus.h>
+#include <boost/bind.hpp>
+
+using boost::lexical_cast;
+
+const unsigned short port = (lunchbox::RNG().get<uint16_t>() % 60000) + 1024;
+
+class Service : public lunchbox::Thread
```
hernando: The fact that the subscriber is contained in a class called Service is really fishy to me.

tests/connection/broker.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    void run() final
+    {
+        zeq::Subscriber subscriber( lunchbox::URI( "foo://127.0.0.1:" +
+                                          lexical_cast< std::string >( port )));
+        BOOST_CHECK( subscriber.registerHandler( zeq::vocabulary::EVENT_ECHO,
+                                        boost::bind( &test::onEchoEvent, _1 )));
+
+        // Using the connection broker in place of zeroconf
+        std::string address = std::string( "127.0.0.1:" ) +
```
hernando: All the IPs being 127.0.0.1 doesn't help to distinguish which is A and which B. Can you use a different variable for each participant? then this code can be used as an example in the documentation.

zeq/connection/broker.h
hernando: I understand that the publisher is connecting to the Broker using a known IP to announce itself. What I don't get is why the broker and the subscriber need to be bound together. To me the broker should be a standalone entity (which could run between networks) responsible of introducing publishers to subscribers, possible taking care of multiple applications (mediating in the case there are collisions in the chosen schemas). I understand that the use case under consideration is simpler, but maybe Broker should be renamed to make it clear that is a simplified version of a more general concept (that would address Bernd's comment somehow). What about BasicBroker
In any case, the publisher-subscriber protocol should be forward compatible in the sense that it should designed with the general broker in mind.

zeq/connection/service.cpp
eile: done

zeq/connection/service.cpp
eile: too

zeq/publisher.h
eile: Future-proof code!! Fixed.

zeq/connection/broker.h
eile: > why the broker and the subscriber need to be bound together 

Because: 

> broker is [...] responsible of introducing publishers to subscribers

and therefore needs to know the subscribers. It is a connection::Broker, and I don't see how it can be made any more complex than it is (in terms of API), or why it should be? It is a first minimal functional implementation and as such can evolve forward and potentially split into different impls later.

tests/connection/broker.cpp
eile: A service subscribes to request events to publish the reply data. What is fishy for you?

tests/connection/broker.cpp
eile: Again, not sure I understand: This is the service side where a broker brokers connection requests for a service. The other side is in test_broker which has a publisher and service::connect().

tribal-tec:
+1

bilgili:
+2

PR #36, opened 2015-01-07
Integrate Lunchbox and FlatBuffers as subprojects

eile:
This leaves the (non-CMake) projects Boost and libzmq as remaining
required dependencies. When they are installed externally, a standard
CMake build setup (without Buildyard) can be used.


hernando:
How is this supposed to work? I don't find where .gitsubprojects is included to be evaluated by CMake

eile:
@hernando:
```
[shichi CMake master]% head -30 SubProject.cmake

# Include this file in a top-level CMakeLists to build several CMake
# subprojects (which may depend on each other).
#
# When included, it will automatically parse a .gitsubprojects file if one is
# present in the same CMake source directory.
# .gitsubprojects contains lines in the form:
# "git_subproject(<project> <giturl> <gittag>)"
#
# All the subprojects will be cloned and configured during the CMake configure
# step thanks to the 'git_subproject(project giturl gittag)' macro
# (also usable separately).
# The latter relies on the add_subproject(name) function to add projects as
# sub directories. See also: cmake command 'add_subdirectory'.
#
# The following targets are created by SubProject.cmake:
# - An 'update_git_subprojects_${PROJECT_NAME}' target to update the <gittag> of
#   all the .gitsubprojects entries to their latest respective origin/master ref
# - A generic 'update' target to execute 'update_git_subrojects' recursively
#
# To be compatible with the SubProject feature, (sub)projects might need to
# adapt their CMake scripts in the following way:
# - CMAKE_BINARY_DIR should be changed to PROJECT_BINARY_DIR
# - CMAKE_SOURCE_DIR should be changed to PROJECT_SOURCE_DIR
#
# A sample project can be found at https://github.com/Eyescale/Collage.git
```

hernando:
Yeah, sorry. The CMake subtree in which I looked into wasn't up to date.

hernando:
+2 (after a quick glance to SubProject.cmake, which I assume has been reviewed elsewhere)

PR #39, opened 2015-02-02
Generic support for JSON serialization

tribal-tec:


tribal-tec:
FYI @eile @hernando @bhentschel @favreau @chevtche @bilgili @jafyvilla

hernando:
This PR doesn't compile for me. I had to add SYSTEM to all include_directories from CMake/FindPackages.cmake, otherwise gcc would complain from a FlatBuffers header.
Then I get an error compiling zeq/detail/vocabulary.cpp:

    /home/jhernando/bbp/Buildyard/src/zeq/zeq/detail/vocabulary.cpp:31:23: error: ‘EVENT_ECHO’ is not a member of ‘zeq::vocabulary’
         zeq::Event event( ::zeq::vocabulary::EVENT_ECHO );

I tried removing all the previous FlatBuffers installation and starting with an empty build directory for zeq, but still the same.

By the way, shouldn't FlatBuffers install the CMake config script under share. It seems that it has to be nested inside zeq to be detected sucessfully.

hernando:
I've found out why it doesn't compile. The subproject commit for FlatBuffers is outdated, it should be edba9db.

hernando:
+2, somebody else to merge

zeq/detail/exit.fbs
```diff
@@ -0,0 +1,6 @@
+namespace zeq.vocabulary;
+
+table Exit {
+}
+
+root_type Exit;
```
eile: I remember seeing this somewhere else... Where does this one come from, or why is it added now?

eile:
+1

zeq/detail/exit.fbs
tribal-tec: As there is no data for that, this schema had no real value, but the code generation uses that to create the according event type.

PR #45, opened 2015-02-23
Added a imageJPEG event.

chevtche:


chevtche:
No. We separated the events because with our current architecture, having a generic imageEvent with the type ( JPEG, PNG, RAW ) in it will create states in Renderers. We want to avoid that.

chevtche:
For the std::vector thingy I am not sure. Maybe I don't understand something but if you look in the
hbp/vocabulary.cpp at lines 40-47 the deserializeVector function contains a ugly for loop. I don't understand how it is possible to avoid copy.

chevtche:
I think that:  vector.assign( flatbufferVec->Data(), flatbufferVec->Data() + flatbufferVec->size() ) should
work. Is ther a copy? I really don't understand why the deserializeVector function is not using that ?

hernando:
I think I confused you.

My comment was only about the API, not the deserialization implementation. The copy the compiler can avoid is from the function return value (rvalue) to the destination variable (lvalue). This must be one of the few cases were the standard allows the compiler to produce code with different semantics for optimized and non-optimized code: http://en.wikipedia.org/wiki/Return_value_optimization.

The code that you suggest is making a copy because std::vector owns the storage, it can never contain a shallow pointer. Which makes me think that the API can't return std::vector if we want to avoid any copy at all in the implementation.

In any case, my point is that ImageJPEG is nothing but a wrapper around a plain C array to include the size. I think there must be a better way to express this than having a class which isn't really JPEG specific.

hernando:
I don't understand why having a single message with an additional parameter for the type adds state to the renderers but having two different messages doesn't.
If you need to send informations A, B and C as a message you can put A, B in the message type and C as an attribute or you can put A in the type and B and C as attributes. I doesn't matter how you do it, you still need to know A, B, C.

chevtche:
In our current architecture we have a zeq Request event. This Request event carries informations about the zeq event that Livre must return. If we have ImageRaw event and ImageJPEG event it is sufficient. But with a generic image event Livre cannot know what image it needs to return, it only knows that it needs to return a generic image event but not the type. To be able to manage this generic image event we either need to change the architecture or introduce states like "imageFormat=JPEG" which we can set with another zeq event like setImageFormat or so.

bilgili:
I am thinking exactly like you Juan on this case. A generic image message. But, to me,  scope of the work related to this is not including this work. The generic image story  (which is unfortunately not forseen/planned)  is much bigger than the small story that depends on this feature. IMHO, lets keep it simple until we have the urge to add another image type.

hernando:
After the last comment from Gregory I understand that the shortcoming comes from the fact that the only parameter for the request message is the type of the message that will be sent as answer. Clearly, this won't be sufficient in the future.

chevtche:
I agree. I also prefere the generic image event. Do you think that adding a getImage event is good in our case. getImage event can have the type of the image in it.

hernando:
That's probably the smallest amount of work that will solve the issue. The generic request is OK when you don't have something else, and since adding generic parameters to that message will easily break message-type safety (moves the burden of checking parameters into the runtime), adding a message for this specific case is fine. I think that more parameters will appear in the long run and that will justify having a separate message.

chevtche:
After a discution with Daniel and Stefan we dicided to remove the RAW image and use only JPEG for now. Later, all the encoding will be moved to a separate library.

chevtche:
Retest this please

chevtche:
Can this be merged?

tribal-tec:
+1

bilgili:
+2

PR #46, opened 2015-03-03
Added a simple keep-alive event

chevtche:


tribal-tec:
Name it EVENT_HEARTBEAT. keepalive would be the action one takes to get more heartbeats

hernando:
I didn't want to be pedantic, but I was thinking the same.

tribal-tec:
Just thinking: the existing EVENT_ECHO would be sufficient, but probably semantically not correct, right?

chevtche:
Echo seems false... There is no echo...

hernando:
Unless you send KEEP_ALIVE from the peer and the receiver answers EVENT_ECHO, but that requires two publishers and subscribers, which is not the simplest solution.

I prefer adding the HEARTBEAT message.

chevtche:
Ok. Updated.

chevtche:
Can this me merged?

hernando:
+2, I can't merge

hernando:
Wait a second, you have to fix the commit message.

chevtche:
Ok. Thank you.

chevtche:
Ok. It's fixed.

PR #55, opened 2015-04-10
Made zeq compatible with third party applications that do not yet support C++11

favreau:


favreau:
Since my first commit, I realized that boost was providing feature based macros that are more suitable than a generic C++11 macro, hence these new changes.

zeq/event.h
```diff
@@ -35,7 +38,15 @@ class Event : public boost::noncopyable
...
 
     /** Move ctor @internal */
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
     ZEQ_API Event( Event&& rhs );
+#else
+    ZEQ_API Event( const Event& rhs )
+    {
+        _impl = rhs._impl;
+        const_cast<Event &>(rhs)._impl = 0;
+    }
```
eile: If possible, I'ld rather make rhs non-const and remove the const-cast. This way it is clear in the API.

zeq/event.h
```diff
@@ -55,11 +66,22 @@ class Event : public boost::noncopyable
     ZEQ_API flatbuffers::Parser& getParser();
     ZEQ_API const flatbuffers::Parser& getParser() const;
 
+#ifdef BOOST_NO_CXX11_RVALUE_REFERENCES
+    Event& operator=( Event& rhs )
+    {
+        _impl = rhs._impl;
+        const_cast<Event &>(rhs)._impl = 0;
```
eile: no const_cast needed.

eile: Move together with C++11 move operator.

zeq/eventDescriptor.h
```diff
@@ -32,7 +32,10 @@ enum EventDirection
  * It contains the REST api name, the 128bit zeq event id and the schema
  * needed for automatic JSON serialization/deserialization.
  */
-struct EventDescriptor : public boost::noncopyable
+struct EventDescriptor
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
+    : public boost::noncopyable
+#endif
```
eile: Should it not stay noncopyable? To my understanding, it is moveable but not copyable.

zeq/eventDescriptor.h
```diff
@@ -46,7 +49,16 @@ struct EventDescriptor : public boost::noncopyable
...
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
     ZEQ_API EventDescriptor( EventDescriptor&& rhs );
+#else
+    ZEQ_API EventDescriptor( const EventDescriptor& rhs )
+    {
+        _impl = rhs._impl;
+        const_cast<EventDescriptor&>(rhs)._impl = 0;
+    }
+
+#endif
```
eile: Switch if/else. Me not loves not double negatives.

zeq/eventDescriptor.h
```diff
@@ -46,7 +49,16 @@ struct EventDescriptor : public boost::noncopyable
                              const EventDirection eventDirection);
 
     /** Move ctor @internal */
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
     ZEQ_API EventDescriptor( EventDescriptor&& rhs );
+#else
+    ZEQ_API EventDescriptor( const EventDescriptor& rhs )
+    {
+        _impl = rhs._impl;
+        const_cast<EventDescriptor&>(rhs)._impl = 0;
```
eile: Same const comment as above.

zeq/event.h
hernando: The correct declaration of an assignment operator is Event& operator=( const Event& rhs )

zeq/event.h
hernando: Maybe you're right, but I wonder if all client code can compile without the const modifier, because what you're suggesting is moving the const_cast from the implementation to the client code.

zeq/eventDescriptor.h
hernando: That's only true if the copy constructor has the non-const overload. However I'm wondering if that can cause problems in client code.

I remember playing with const and non-const copy constructors in the past and I remember having problems, but I can't tell what exactly right now.

Another alternative suggested at the bottom of [this page](http://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Move_Constructor), suggests to use a mutable member to pass the ownership from rhs. This solution needs the class to be copyable and allows data sharing between the multiple copies but doesn't need any const cast. I actually prefer this solution, despite being less safe, because we should consider the C++03 support a legacy feature.

zeq/event.h
favreau: I'd rather leave the const in both the constructor and the assignment operator. As mentioned by Juan, I also believe this is the correct declaration. No one is really happy with the const_cast, but we can admit this is a temporary solution for compilers that do not support C++11.

zeq/eventDescriptor.h
favreau: noncopyable can actually be kept, and is valid in the context of the C++03 implementation. I'm happy to set it back.

zeq/eventDescriptor.h
favreau: Good point. Done.

zeq/event.h
eile: @hernando: No, the client code should not const-cast, since this is effectively a move operator, not an assignment operator.

eile: @favreau / @hernando : Yes, it is the correct(er) declaration, but then the code should not modify rhs. Implementing move semantics in a copy operation will lead to nasty bugs. Make it explicit or let it be.

zeq/event.cpp
```diff
@@ -13,49 +13,64 @@ Event::Event( const uint128_t& type )
...
 
 Event::~Event()
 {
     delete _impl;
 }
 
+void Event::copy( const Event& rhs )
+{
+    _impl = new detail::Event( rhs.getType() );
+    LBASSERT(_impl);
```
tribal-tec: do we need all the asserts? if new fails, bad_alloc is thrown.

zeq/event.h
```diff
@@ -35,7 +35,20 @@ class Event : public boost::noncopyable
     ZEQ_API explicit Event( const uint128_t& type );
 
     /** Move ctor @internal */
```
tribal-tec: comment does not match, it's copy and assignment here, not move

zeq/eventDescriptor.h
```diff
@@ -40,13 +40,26 @@ struct EventDescriptor : public boost::noncopyable
      * @param eventType the zeq event's uint128
      * @param schema the flatbuffers schema as string
      */
-    ZEQ_API EventDescriptor( const std::string& restName,
-                             const uint128_t& eventType,
-                             const std::string& schema,
-                             const EventDirection eventDirection);
+    ZEQ_API explicit EventDescriptor( const std::string& restName,
```
tribal-tec: explicit is useless for multi argument ctors

zeq/event.cpp
```diff
@@ -13,49 +13,64 @@ Event::Event( const uint128_t& type )
...
+#endif
 
 Event::~Event()
 {
     delete _impl;
 }
 
+void Event::copy( const Event& rhs )
+{
+    _impl = new detail::Event( rhs.getType() );
```
tribal-tec: either here for all or in each invocation from the assignment operator should be a check for self assignment.

eile: + memleak of old _impl

zeq/eventDescriptor.cpp
```diff
@@ -29,22 +32,33 @@ EventDescriptor::~EventDescriptor()
...
 EventDirection EventDescriptor::getEventDirection() const
 {
+    LBASSERT(_impl);
     return _impl->eventDirection;
 }
 
+void EventDescriptor::copy( const EventDescriptor& rhs )
+{
+    _impl = new detail::EventDescriptor(
+        rhs.getRestName(), rhs.getEventType(), rhs.getSchema(), rhs.getEventDirection() );
```
eile: idem

zeq/event.cpp
favreau: It was a check that I thought was safer, but you made a point with the bad_alloc. I'll remove the asserts.

tribal-tec:
+1

hernando:
After the decision taken in the workshop last week this review has to be abandoned.

PR #57, opened 2015-04-20
Superproject fixes

eile:


eile:
@hernando Please review

CMakeLists.txt
```diff
@@ -4,9 +4,8 @@
 cmake_minimum_required(VERSION 2.8)
 project(zeq)
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMake
-  ${PROJECT_SOURCE_DIR}/CMake/common)
+  ${CMAKE_SOURCE_DIR}/CMake/common)
```
hernando: Could you exaplain me why it has to be the top level CMake/common (I guess the superproject one)?
Does this mean all subprojects must be prepared to use the same CMake/common commit to work?

eile: Yes, this is correct. The alternative currently in place is almost unusable when you need to change something in CMake/common, since you have to copy the file everywhere.

tests/connection/broker.cpp
```diff
@@ -96,7 +96,8 @@ class NamedSubscriber : public Subscriber
                 return BrokerPtr(
                     new zeq::connection::Broker(
                       "zeq::connection::test_named_broker", subscriber, mode ));
-                lunchbox::sleep( 100 );
+                // cppcheck-suppress unreachableCode
```
tribal-tec: cppcheck is right here, no? there is a return statement before, so why would that 'fail'?

hernando: You're right. I think the sleep call was intended to be inside the catch clause.

eile: doh! Will fix.

hernando:
+1

eile:
Retest this please

bilgili:
+1

PR #63, opened 2015-05-01
Replace boost by C++11 std equivalents; boost unit_test_framework remains

tribal-tec:


hernando:
Have you considered *using namespace std::placeholders*?

tribal-tec:
Not really, but might be the better solution. I'm kind of allergic to using namespace.

zeq/subscriber.cpp
```diff
@@ -51,7 +49,7 @@ class Subscriber
 
     ~Subscriber()
     {
-        BOOST_FOREACH( SocketType socket, _subscribers )
+        for( const SocketType& socket : _subscribers )
```
eile: const ref for an enum?

eile:
+1

zeq/subscriber.cpp
tribal-tec: typedef std::pair< std::string, void* > SocketType;

eile: ok, strange name then. SocketInfo? hmm, not so much better.

hernando: If this is the key pair type used in a map, I think it's preferable to use MapType::mapped_type (not sure about the spelling) or even auto (I think auto works here)

tribal-tec:
Updated

tribal-tec:
ping

tribal-tec:
ping

tribal-tec:
retest this please

tribal-tec:
^ was about to test the build status update of the PR. seems to work :)

tribal-tec:
retest this please

bilgili:
+2

PR #65, opened 2015-05-07
Fixed building error due to std::to_string not supporting short unsigned int

juanjosegarcan:
Fixed building error due to std::to_string not supporting short unsigned int


hernando:
The current master should be fixing all these errors, are you using it?

PR #66, opened 2015-05-07
Fix MSVC build

tribal-tec:
Fixes the error caused by warning C4521: 'zeq::Receiver' : multiple copy
constructors specified


hernando:
+1

PR #69, opened 2015-05-22
Replaced Lunchbox by Servus.

hernando:
Some other minor changes


hernando:
Depends on HBPVIS/Servus#3.

.gitignore
```diff
@@ -4,5 +4,5 @@ Makefile
 CMakeLists.txt.user*
 CMake/common/
 ___*
-Lunchbox/
+servus/
```
eile: Capital

CMakeLists.txt
```diff
@@ -16,7 +16,6 @@ include(Common)
 
 set(COMMON_PROJECT_DOMAIN eu.humanbrainproject)
 set(ZEQ_DESCRIPTION "ZeroEQ - Zero Event Queue")
-set(ZEQ_DEPENDENT_LIBRARIES Lunchbox)
```
eile: Iirc Servus leaks through the API, so should be used here.

tests/broker.h
```diff
@@ -1,23 +1,51 @@
 
-/* Copyright (c) 2014, Human Brain Project
+/* Copyright (c) 2015, Human Brain Project
```
eile: 2014-2015?

tests/broker.h
```diff
@@ -1,23 +1,51 @@
...
 {
-lunchbox::URI buildURI( const std::string& hostname = std::string( ))
+
+servus::URI buildPublisherURI( const std::string& schema )
+{
+    return servus::URI( schema + ":" );
+}
+
+servus::URI buildPublisherURI( const std::string& schema,
+                                 const unsigned int port )
```
eile: indent

AUTHORS.txt
```diff
@@ -2,4 +2,9 @@ The following people have contributed to zeq:
 
 -------- git master -------
 
+Stefan Eilemann
```
tribal-tec: sort alphabetically

CMake/zeq.cmake
```diff
@@ -2,7 +2,7 @@
 set(ZEQ_VERSION 0.1)
 set(ZEQ_REPO_URL https://github.com/HBPVIS/zeq.git)
 set(ZEQ_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/src/zeq")
-set(ZEQ_DEPENDS hbpvis REQUIRED libzmq FlatBuffers Boost Lunchbox)
+set(ZEQ_DEPENDS hbpvis REQUIRED libzmq FlatBuffers Servus OPTIONAL Boost)
```
tribal-tec: optional unit tests is a bit fishy to me...

zeq/connection/broker.cpp
```diff
@@ -4,12 +4,16 @@
  */
 
 #include "broker.h"
+
 #include <zeq/detail/port.h>
 #include <zeq/detail/socket.h>
 #include <zeq/receiver.h>
+#include <zeq/log.h>
+
+#ifdef ZEQ_USE_LUNCHBOX
```
tribal-tec: that's wrong

zeq/detail/event.h
```diff
@@ -58,8 +77,8 @@ class Event
     /** store the serialized data in here */
     flatbuffers::Parser parser;
 
-    /** setData() uses this instead of fbb during deserialization */
```
tribal-tec: keep the comment please

zeq/event.h
```diff
@@ -60,6 +60,7 @@ class Event
 
     friend class detail::Subscriber;
     void setData( const void* data, const size_t size );
+    void setData( const ConstByteArray& data, const size_t size );
```
tribal-tec: the upper is not used the anymore? Or why we need two setData() now?

CMakeLists.txt
hernando: That's right. The prototype with out Lunchbox at all didn't have it, but now the URI is leaked.

zeq/event.h
hernando: Both are used. The difference was documented in the implementation class.

zeq/event.cpp
```diff
@@ -59,4 +54,14 @@ const flatbuffers::Parser& Event::getParser() const
     return _impl->parser;
 }
 
+void Event::setData( const void* data, const size_t size )
+{
+    _impl->setData( data, size );
+}
+
+void Event::setData( const ConstByteArray& data, const size_t size )
```
tribal-tec: I still don't get why we need two setData functions?

hernando: Because one takes the ownership of the pointer (through a shared pointer) and the other one doesn't. I moved the documentation to zeq/event.h hoping it would be clearer.
An alternative is to provide just the smart pointer version and let the caller decide by providing a custom deallocator.

tribal-tec: The caller is the library, no user code should call that. So one version is for serialize, the other for deserialize?

hernando: Actually I don't know, maybe one of them is not used at all. I'll double check.

tribal-tec:
lcov would tell you :) Have you checked that coverage? I'd like not to drop if possible. See numbers in commit message from https://github.com/HBPVIS/zeq/commit/e281cb9dad8c2a00913f23decc3f73ca74277c56

hernando:
The pubSub test (test_publish_blocking_receive_zeroconf) is intermitent in Lugano, mostly failing, but it works in Ubuntu 14.04. Has this been noticed before or is it due to this PR?

To give more details, the test prints this:

    /gpfs/bbp.cscs.ch/home/hernando/Buildyard/src/Servus/servus/avahi/servus.h:359: Resolver error: Timeout reached
    /gpfs/bbp.cscs.ch/home/hernando/Buildyard/src/Servus/servus/avahi/servus.h:359: Resolver error: Timeout reached
    Publisher giving up after 20s
    terminate called after throwing an instance of 'std::runtime_error'
      what():  Publisher giving up after 20s
    unknown location(0): fatal error in "test_publish_blocking_receive_zeroconf": signal: SIGABRT (application abort requested)

However this is not very meaningful because when it works the output looks like this:

    /gpfs/bbp.cscs.ch/home/hernando/Buildyard/src/Servus/servus/avahi/servus.h:359: Resolver error: Timeout reached
    /gpfs/bbp.cscs.ch/home/hernando/Buildyard/src/Servus/servus/avahi/servus.h:359: Resolver error: Timeout reached
    Subscribed to tcp://bbpviz1.cscs.ch:51264
    
    *** No errors detected

hernando:
The code coverage has dropped drastically and right now it's roughly 75% (not due to this PR)

tribal-tec:
Ok, bad for the coverage, but not relevant for this PR then. The test is green with current master, I never noticed it before. Is it deterministically reproducible?

zeq/event.h
```diff
@@ -59,7 +59,9 @@ class Event
     Event& operator=( const Event& ) = delete;
 
     friend class detail::Subscriber;
-    void setData( const void* data, const size_t size );
+
+    /** Set the a raw buffer as event data. */
```
tribal-tec: grammar: Set the a raw... 
Besides, it's a private function, so I would not put a documentation for it, unless you see value for it.

hernando: It's used from another friend class, since it's already there it doesn't harm.

hernando:
The intermitent test is not deterministic, if fails randomly about half of the times.

tribal-tec:
retest this please

PR #70, opened 2015-05-26
Relax and document thread-safety for JSON parsing

eile:
Stumbled upon this when creating independent events in parallel: When
their types are not registered, the (unneeded) schema parsing in the
event ctor called in parallel unordered_map::operator [] which is not
thread safe.

Moved schema parsing to JSON code and made getSchema() "const"

### DIFF
--- a/zeq/detail/event.h
+++ b/zeq/detail/event.h
@@ -1,12 +1,10 @@
 
-/* Copyright (c) 2014-2015, Human Brain Project
- *                          Daniel Nachbaur <daniel.nachbaur@epfl.ch>
+/* Copyright (c) 2014, Human Brain Project
+ *                     Daniel Nachbaur <daniel.nachbaur@epfl.ch>
  */
 
 #ifndef ZEQ_DETAIL_EVENT_H
 #define ZEQ_DETAIL_EVENT_H
-
-#include "vocabulary.h"
 
 #include <zeq/types.h>
 #include <zeq/event.h>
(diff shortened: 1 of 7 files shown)


hernando:
Typo on commit message "mad"

zeq/vocabulary.h
```diff
@@ -59,7 +61,22 @@ ZEQ_API Event serializeRequest( const uint128_t& eventType );
  * @return an uint128_t to identify the zeq event that should be published.
  */
 ZEQ_API uint128_t deserializeRequest( const Event& event );
+//@}
 
+/**
+ * @name JSON/binary event translation.
+ *
+ * These functions are not thread-safe, that is, calling registerEvent()
```
tribal-tec: You call that function in user code? Where and why? Usually, flatbuffers generated calls that on init, so I'm wondering what use case I'm missing.

zeq/detail/event.h
```diff
@@ -25,13 +23,7 @@ class Event
...
     Event( const uint128_t& type_ )
         : type( type_ )
-    {
-        const std::string& schema = vocabulary::detail::getSchema( type );
-
-        // populate parser with schema from type
-        if( !schema.empty() && !parser.Parse( schema.c_str( )))
-            LBTHROW( std::runtime_error( parser.error_ ));
-    }
+    {}
```
tribal-tec: that change in behavior will affect RESTBridge if I'm not mistaken. Why the change?

tribal-tec:
@chevtche @favreau  Can you check if this ZEQ version would work with RESTBridge?

tribal-tec:
Document exceptions accordingly for Event ctor and (de)serializeJSON functions.
+1 then.

PR #72, opened 2015-05-29
Fix MSVC build

tribal-tec:
FYI @hernando

### DIFF
--- a/zeq/detail/port.h
+++ b/zeq/detail/port.h
@@ -8,6 +8,7 @@
 #ifndef ZEQ_DETAIL_PORT_H
 #define ZEQ_DETAIL_PORT_H
 
+#include <zeq/api.h>
 #include <zeq/types.h>
 
 namespace zeq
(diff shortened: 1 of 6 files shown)


zeq/detail/port.h
```diff
@@ -19,7 +20,7 @@ namespace detail
 uint16_t getPort( const std::string& name );
 
 /** @return a random port number. */
-uint16_t getRandomPort();
+ZEQ_API uint16_t getRandomPort();
```
hernando: I know it's already merged, but this looks fishy, doesn't it? A function in zeq/detail shouldn't be needed to be exported.

tribal-tec: It's used in all the tests, hence the export is needed there. The detail is not installed, so nobody outside of ZeroEQ can use it. It's more fishy that this function is used in the tests or even that ZeroEQ provides a random port generation function :)

hernando: I agree with that. It should me moved to a common header for tests, unless used in the implementation, which could be the case.

PR #74, opened 2015-06-23
Fix exception with broken DNS and zeroconf

eile:


eile:
@bilgili FYI in case I crashed your Livre just now...

tribal-tec:
+1

PR #75, opened 2015-06-26
Event for frame setting and animation

eile:


zeq/hbp/detail/frame.fbs
```diff
@@ -0,0 +1,16 @@
...
+namespace zeq.hbp;
+
+table Frame
+{
+    /// starting frame number for an animation loop
+  start:uint;
+    /// current frame (wrapped into [start end])
+  current:uint;
+    /// end frame number for an animation loop
+  end:uint;
```
tribal-tec: add delta to have different increment than 1 (could even be decrement, like -2 to animate backwards 2 times the 'normal' speed)?

tribal-tec:
+1 otherwise. I would have put it into livre for now, but is generic enough actually to put in in hbp vocabulary.

zeq/hbp/detail/frame.fbs
eile: will do.

hernando:
Maybe you've already discussed it but why not using time instead of frames?
Or do you think that can be provided as a separate event and applications can choose which one to use?

eile:
Updated

eile:
We did discuss it, and while I was intuitively for float time as well, the argument was made for Livre to use frame numbers, translated to time by a data-source dependent impl.

hernando:
Is Livre requiring frame numbers related to the encoding of the time dimension in the node id?
If that's the case, isn't that an artificial requirement imposed by an implementation detail?

bilgili:
+2

eile:
> Is Livre requiring frame numbers related to the encoding of the time dimension in the node id?

Yes.

> If that's the case, isn't that an artificial requirement imposed by an implementation detail?

Yes, but so would be a float time. The time interpretation again would be up to the data source. Both make sense, and if you think about it a lot of software works in concrete timesteps/frames, for example in movie editing apps.

hernando:
Yes, but in movie production the conversion from frames to timestamps is done according to a well defined standard and there are only a handful of standard to choose from.
What makes more sense depends on whether your consider the event part of a closed application or it can be used to communicate applications that may have very different notions of how frame numbers are translated into timestamps. For example, to do visualization of multi-scale simulations, it's probably better to always work with timestamps.

eile:
...and then we need a scaling as for the camera :P

I should not be the only one in this discussion, so we should find another medium to discuss.

PR #76, opened 2015-07-06
Release 0.3 preparations

tribal-tec:


README.md
```diff
@@ -1,4 +1,43 @@
...
+environments are tested:
+
+* Linux: Ubuntu 14.04, RHEL 6 using gcc 4.8
+* Windows: 8 using Visual Studio 12
+* Mac OS X: 10.9 and 10.10 using clang 6
+
+ZeroEQ requires the following external, pre-installed dependencies:
+
+* ZeroMQ 4.0 or later
+* Boost (tested with 1.54)
```
hernando: Only for unit tests, right?

tribal-tec: True, could/should be even optional

tribal-tec:
ping

CMakeLists.txt
```diff
@@ -24,5 +24,7 @@ add_subdirectory(zeq)
 add_subdirectory(tests)
 add_subdirectory(tools)
 
+set(DOXYGEN_MAINPAGE_MD README.md)
+set(DOXYGEN_EXTRA_INPUT ${PROJECT_SOURCE_DIR}/README.md)
```
hernando: README.md twice?

tribal-tec: that's how it works afaiu?!?

eile:
+1

README.md
```diff
@@ -1,4 +1,43 @@
...
+* Linux: Ubuntu 14.04, RHEL 6 using gcc 4.8
+* Windows: 8 using Visual Studio 12
+* Mac OS X: 10.9 and 10.10 using clang 6
+
+ZeroEQ requires the following external, pre-installed dependencies:
+
+* ZeroMQ 4.0 or later
+* Boost (tested with 1.54) for unit tests
+
+~~~
```
hernando: Doxygen doesn't understand this syntax. You can use four spaces to indicate a verbatim block but I'm not sure Github understands that.

doc/Changelog.md
```diff
@@ -0,0 +1,28 @@
+# Changelog
```
hernando: Isn't it better:

    Chagelog {#Changelog}
    ========

README.md
tribal-tec: My doxygen does

doc/Changelog.md
tribal-tec: Why?

CMakeLists.txt
hernando: I see. DOXYGEN_EXTRA_INPUT is needed because README.md is not under doc. If all documentation files were under doc no extra variable would be needed. DOXYGEN_MAINPAGE_MD is not needed if you add a title with {#mainpage} to README.md. The funny thing is that doxygen adds a title automatically to the main page if you omit it, so if you have removed it because you think it'd look nicer it won't work as expected.
I'm realizing that we have dozens of variations on how we do things, and it gets confusing.

doc/Changelog.md
hernando: Let me check if your alternative shows correctly in doxygen.

README.md
hernando: I removed the comment because I double checked. I remember seeing this failing, but I can't recall the conditions.

doc/Changelog.md
hernando: If you do this you can give an arbitrary title to the page. Otherwise, doxygen names it after the filename. So for the change log is not needed.

hernando:
+1, feel free to merge

PR #78, opened 2015-07-14
CMake update after Eyescale/CMake#341

tribal-tec:


CMakeLists.txt
```diff
@@ -18,7 +18,12 @@ set(COMMON_PROJECT_DOMAIN eu.humanbrainproject)
 set(ZEQ_DESCRIPTION "ZeroEQ - Zero Event Queue")
 set(ZEQ_DEPENDENT_LIBRARIES Servus)
 
-include(FindPackages) # Generated by Buildyard and checked in
+common_package(Boost REQUIRED COMPONENTS unit_test_framework)
```
hernando: Shouldn't we make boost optional?

PR #81, opened 2015-08-03
Implement (optional) ZeroBuf support

tribal-tec:


tribal-tec:
retest this please

tests/connection/broker.cpp
```diff
@@ -52,7 +56,11 @@ class Subscriber
...
         {
             if( subscriber.receive( 100 ))
-                received = true;
+            {
+#ifdef ZEQ_USE_ZEROBUF
+                received = echo.gotData && received;
+#endif
+            }
         }
     }
```
eile: Shouldn't this be two separate tests? Right now the FB code path is not tests as far as I understand.

tests/connection/broker.cpp
```diff
@@ -70,6 +78,12 @@ class Subscriber
     mutable std::mutex _mutex;
     bool _started;
 
+    void onEchoEvent( const zeq::Event& event )
+    {
+        test::onEchoEvent( event );
+        received = true;
+    }
+
```
eile: Ah, here. hmm.

zeq/subscriber.h
```diff
@@ -73,6 +73,27 @@ class Subscriber : public Receiver
      */
     ZEQ_API bool deregisterHandler( const uint128_t& event );
 
+    /**
+     * Subscribe a ZeroBuf object to updates published by any connected
+     * receiver.
```
eile: Subscribe a ZeroBuf object to receive updates from any connected publisher

zeq/subscriber.h
```diff
@@ -73,6 +73,27 @@ class Subscriber : public Receiver
...
+     *
+     * Every received update will directly applied on the ZeroBuf object. To
```
eile: Every update will be directly applied on the ZeroBuf object during receive().

eile: The subscribed object instance has to be valid until unsubscribe().

eile:
+1 otherwise.

PR #85, opened 2015-08-11
Added animation event

favreau:


zeq/hbp/detail/animation.fbs
```diff
@@ -0,0 +1,14 @@
...
+//                     Cyrille Favreau <cyrille.favreau@epfl.ch>
+//
+namespace zeq.hbp;
+
+enum AmimationAction : ubyte { Pause, Play, Backwards }
+
+table Animation
+{
+  action: AmimationAction;
+  frame : uint; // frame number
```
eile: This is already part of frame.fbs - maybe we should simply update this event with the enum?

eile: Plus, not the enum. It is an int in livre, so it should be here one too.

eile:
-1

eile:
-2 in fact, animation delta is already part for frame.fbs. I don't see any need for this PR.

favreau:
We actually had a discussion before end about defining a new event or not. IMHO, both events do not deal with the same features. The frame event transports information about the volume + cache and should be not used for animation purposes. Using a unique event will lead to loops where, for example, Fivox will publish a current frame that will then be used by Livre to render the frame, and we do not want this to happen, hence the creation of animation dedicated event.

eile:
I don't understand. There are now volume or cache parameters in frame.fbs:
```
    /// starting frame number for an animation loop
  start:uint;
    /// current frame (wrapped into [start end])
  current:uint;
    /// end frame number for an animation loop
  end:uint;
    /// animation delta between start<->end
  delta:int;
```

Re loops: That's an interesting problem, but should not be solved by having two events for the same thing. But why would Fivox publish this event?

PR #94, opened 2015-08-27
Added CellSetBinaryOp to HBP vocabulary

segalindo:
This includes the CellSetBinaryOp serialization into the HBP vocabulary, consisting of a pair of unsigned integer vectors and the specified operation.

### DIFF
--- a/tests/hbp/serialization.cpp
+++ b/tests/hbp/serialization.cpp
@@ -95,3 +95,34 @@ BOOST_AUTO_TEST_CASE( imageJPEGEvent )
                                    deserializedImage.getDataPtr(),
                                    deserializedImage.getDataPtr() + size );
 }
+
+BOOST_AUTO_TEST_CASE( cellSetBinaryOp )
+{
+  zeq::hbp::data::CellSetBinaryOp cellSet ({ 0, 2, 4, 6 },
+                                           { 1, 3, 5, 7 },
+                                           zeq::hbp::CellSetOpType::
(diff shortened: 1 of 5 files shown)


tests/hbp/serialization.cpp
```diff
@@ -95,3 +95,31 @@ BOOST_AUTO_TEST_CASE( imageJPEGEvent )
                                    deserializedImage.getDataPtr(),
                                    deserializedImage.getDataPtr() + size );
 }
+
+BOOST_AUTO_TEST_CASE( binarySetOperation )
+{
+  unsigned int firstSet[ ] = { 0, 2, 4, 6 },
+               secondSet[ ] = { 1, 3, 5, 7 };
+  const std::vector< unsigned int  > first (
```
hernando: Are we allowing initialization lists from C++11? If that's the case use them.

zeq/hbp/vocabulary.h
```diff
@@ -69,6 +71,17 @@ struct ImageJPEG
     const uint8_t* _data;
 };
 
+
+
+struct CellSetBinaryOp
```
hernando: Provide some minimal doxygen documentation (e.g. the event related to this class, the purpose, ...)

zeq/hbp/vocabulary.h
```diff
@@ -69,6 +71,17 @@ struct ImageJPEG
...
+{
+public:
+
```
hernando: Add a constructor with the three parameters.

hernando:
+1

PR #95, opened 2015-08-31
Removing the enum for cell set operations from cellSetBinaryOp.fbs

hernando:
This was requiring to include the flatbuffers header in the vocabulary header,
which in turn makes Flatbuffers leak. The enum has been moved to the C++ side.
The previous enum was not kept to avoid a redundancy.


ptoharia:
+1

PR #97, opened 2015-09-16
Added a small command line tool to send events parsed from a script file

hernando:
The input file contains the event list with the event type, the pause to
wait before sending the event and the event parameters. The only events
supported at the moment are zeq.hbp.ToggleIdRequest and
zeq.hbp.CellSetBinaryOp. More events will be added as needed.


hernando:
FYI @EagleSixty6, @ptoharia

doc/Changelog.md
```diff
@@ -6,6 +6,12 @@
   Implement (optional) ZeroBuf support
 * [94](https://github.com/HBPVIS/zeq/pull/94):
   Added CellSetBinaryOp to HBP vocabulary
+* [97](https://github.com/HBPVIS/zeq/pull/97):
+  Added a small command line tool to send events parsed from a script file.
+  The input file contains the event list with the event type, the pause to
+  wait before sending the event and the event parameters. The only events
+  supported at the moment are zeq.hbp.ToggleIdRequest and
+  zeq.hbp.CellSetBinaryOp. More events will be added as needed.
```
eile: Too long for a changelog

tools/eventGenerator.cpp
```diff
@@ -0,0 +1,229 @@
+
+/* Copyright (c) 2014-2015, Human Brain Project
```
eile: Time travel?

tools/eventGenerator.cpp
```diff
@@ -0,0 +1,229 @@
...
+ *                          Juan Hernando <jhernando@fi.upm.es>
+ */
+
+#include <zeq/zeq.h>
+#include <zeq/hbp/hbp.h>
+#include <zeq/hbp/hbp.h>
```
eile: twice?

tools/eventGenerator.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <servus/uri.h>
+
+#include <chrono>
+#include <thread>
+#include <algorithm>
+#include <fstream>
+#include <string.h>
+
+const char* scriptFile = 0;
+const char* uri = "hbp://";
```
eile: why not std::string?

tools/eventGenerator.cpp
```diff
@@ -0,0 +1,229 @@
...
+            {
+                std::cerr << "Missing parameter to schema" << std::endl;
+                exit( -1 );
+            }
+        }
+        if( strcmp( argv[i], "--help" ) == 0 || strcmp( argv[i], "-h" ) == 0 )
+        {
+            printUsageAndExit( argv[0], 0 );
+        }
+        else if( !scriptFile )
```
eile: I guess no else needed since exit()?

Also, swap if/else blocks to remove negation.

tools/event_script_sample
```diff
@@ -0,0 +1,28 @@
...
+
+TOGGLEIDREQUEST 0.1
+201 202 203 204 205 206 207 208 209 210 211 212 213 214 215 216 217 218 219 220 221 222 223 224 225 226 227 228 229 230 231 232 233 234 235 236 237 238 239 240 241 242 243 244 245 246 247 248 249 250 251 252 253 254 255 256 257 258 259 260 261 262 263 264 265 266 267 268 269 270 271 272 273 274 275 276 277 278 279 280 281 282 283 284 285 286 287 288 289 290 291 292 293 294 295 296 297 298 299 300
+
+CELLSETBINARYOP
+10 20 30 40 50 60 70 80 90 100 110 120 130 140 150 160 170 180 190 200 210 220 230 240 250 260 270 280 290 300 310 320 330 340 350 360 370 380 390 400 410 420 430 440 450 460 470 480 490 500
+510 520 530 540 550 560 570 580 590 600 610 620 630 640 650 660 670 680 690 700 701 711 721 731 741 751 761 771 781 791 801 811 821 831 841 851 861 871 881 891 901 911 921 931 941 951 961 971 981 991
+SYNAPTIC_PROJECTIONS
```
eile: This file format should be documented in the tool help (at least).

tools/eventGenerator.cpp
hernando: Not really. The code is intended to do the following: each iteration it parses a command line option. If the string found is not an option then it's considered a positional parameter. The first positional parameter is the script file, if another one appears, it's an error. Even if I swap the blocks the else must remain.
Maybe what confuses you is that this code has two bugs. The second if should be an else if and the assignment of the string to uri is missing.

And this is why I hate coding command line option parsing. But without adding a dependency I didn't have other option.

hernando:
I did a bad push and the review closed itself. I'll have to create a new one because I cannot reopen it.

hernando:
Actually I don't know what happened, but the PR is now at HBPVIS/zeq/master !? I'm fully sure I did not push to origin. I think all the comments had been addressed and the master history looks good, so I'll leave it closed.

PR #101, opened 2015-10-15
Improved project info

eile:


CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ include(Common)
 
 set(COMMON_PROJECT_DOMAIN eu.humanbrainproject)
 set(ZEQ_DESCRIPTION "ZeroEQ - Zero Event Queue")
+set(ZEQ_MAINTAINER "Blue Brain Project <bbp-open-source@googlegroups.com>")
```
tribal-tec: we have an HBP mailing list we could use?

eile: Done, was missing in https://github.com/HBPVIS

PR #103, opened 2015-10-21
Add support for host:0 in URIs (OS-chosen port)

jafyvilla:
zmq expects host:*

### DIFF
--- a/zeq/detail/broker.h
+++ b/zeq/detail/broker.h
@@ -14,11 +14,12 @@ std::string buildZmqURI( std::string host, const uint16_t port )
 {
     if( host.empty( ))
         host = "*";
-    if( host != "*" && port == 0 ) // zmq does not support host:0
-        throw std::runtime_error(
-            "OS-chosen port not supported with hostname " + host );
 
-    return std::string( "tcp://" ) + host + ":" + std::to_string( int( port ));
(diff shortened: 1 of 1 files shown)


eile:
But tcp://:0 works?

zeq/detail/broker.h
```diff
@@ -14,11 +14,12 @@ std::string buildZmqURI( std::string host, const uint16_t port )
...
     if( host.empty( ))
         host = "*";
-    if( host != "*" && port == 0 ) // zmq does not support host:0
-        throw std::runtime_error(
-            "OS-chosen port not supported with hostname " + host );
 
-    return std::string( "tcp://" ) + host + ":" + std::to_string( int( port ));
+    const std::string tcpURI( "tcp://" + host );
+    if( host != "*" && port == 0 ) // zmq expects host:* instead of host:0
+        return tcpURI + ":*";
```
eile: Does tcp://:* also work? If yes, can be simplified to if( port == 0 )

jafyvilla: Mmmm... should the empty host (tcp://:*) work? I don't think it does, but tcp://*:* works fine as far as I can tell

hernando:
AFAIR, tcp://:0 is not a valid URI and it will make Servus throw.

PR #105, opened 2015-11-02
Add Sessions.md

tribal-tec:

PR #109, opened 2015-11-11
Implement session filtering from Sessions.md RFC 

tribal-tec:


zeq/publisher.h
```diff
@@ -17,17 +19,70 @@ namespace detail { class Publisher; }
 /**
  * Serves and publishes events, consumed by Subscriber.
  *
+ * The session is tied to ZeroConf announcment and can be disabled with
```
hernando: typo: announcement

hernando:
The only change you've made to the user docs is the Changelog, at least there should be an update on README.md for the new feature, but actually I'm starting to miss a tutorial for the library.
I've only skimmed through the changes but I'm missing a definition of what a *session* is. It might be obvious for you after having worked on it for weeks, but not to a person only seeing this commit. The terseness of the commit message doesn't help reviewing the code either.

tribal-tec:
For starters, there is this open PR where I was waiting for more comments/input, but I'll merge today due to no more comments:
https://github.com/HBPVIS/zeq/pull/105

I check README.md, but the constructors should have enough information imo. The commit message could be extended, but on the other hand I don't like 20 paragraph messages like you do :) Maybe we can meet in the middle.

hernando:
I write 20 paragraph messages because I can discuss reviews face-to-face and I feel people would struggle even more with some of my reviews if I didn't. Anybody is free to ignore what's written, but you can't make up information.

I'll give a read to #105 now.

hernando:
Maybe you can link to #105 in this commit message for future reference.

tribal-tec:
Ping

zeq/uri.h
```diff
@@ -0,0 +1,58 @@
...
+
+    /** Create URI from string, set schema to "tcp" if empty */
+    ZEQ_API explicit URI( const std::string& uri );
+
+    /** Create URI from string, set schema to "tcp" if empty */
+    ZEQ_API explicit URI( const char* uri );
+
+    /**
+     * Convert from servus::URI, set schema to "tcp" if empty
+     * @todo make this explicit one zeq::Publisher::getURI() returns zeq::URI
```
hernando: one -> once

zeq/detail/defines.h
```diff
@@ -0,0 +1,27 @@
+
```
hernando: rename file to constants.h?

zeq/publisher.cpp
```diff
@@ -16,21 +17,55 @@
...
+#elif __APPLE__
+    char result[PATH_MAX+1];
+    uint32_t size = sizeof(result);
+    if( _NSGetExecutablePath( result, &size ) != 0 )
+        return std::string();
+    const std::string execPath( result );
+#else
+    char result[PATH_MAX];
+    const ssize_t count = readlink( "/proc/self/exe", result, PATH_MAX );
+    const std::string execPath( result, count > 0 ? count : 0 );
```
hernando: print a warning if count < 0?

zeq/publisher.cpp
```diff
@@ -16,21 +17,55 @@
...
+
+    size_t lastSeparator = 0;
+    const size_t length = execPath.length();
+
+    for( size_t i = 0; i < length; ++i )
+        if( execPath[ i ] == '/' || execPath[i] == '\\' )
+            lastSeparator = i+1;
+
+    return lastSeparator == 0 ? execPath :
+                                execPath.substr( lastSeparator, length );
```
hernando: substr takes position and count not two positions, this will access out-of-bounds memory, use substr( lastSeparator ), that will go until the end.

zeq/publisher.cpp
```diff
@@ -16,21 +17,55 @@
...
+    const ssize_t count = readlink( "/proc/self/exe", result, PATH_MAX );
+    const std::string execPath( result, count > 0 ? count : 0 );
+#endif
+
+    size_t lastSeparator = 0;
+    const size_t length = execPath.length();
+
+    for( size_t i = 0; i < length; ++i )
+        if( execPath[ i ] == '/' || execPath[i] == '\\' )
+            lastSeparator = i+1;
```
hernando: use find_last_of and I'd have different codepaths for the separator for Windows and the rest of the world.

zeq/publisher.cpp
tribal-tec: If your UNIX can resolve the path your application, what should the user then?

zeq/publisher.cpp
tribal-tec: The whole code was C&P from Lunchbox. I suggest that you fix it there and I use the fixed version here :)

tests/uri.cpp
```diff
@@ -0,0 +1,97 @@
...
+    BOOST_CHECK_EQUAL( copy, zeq::URI( servus::URI( "inproc://" )));
+
+    copy = servus::URI( "host" );
+    BOOST_CHECK_EQUAL( copy.getScheme(), "tcp" );
+    BOOST_CHECK_EQUAL( copy, zeq::URI( servus::URI( "host" )));
+    BOOST_CHECK_EQUAL( copy, servus::URI( "host" ));
+    BOOST_CHECK( copy != servus::URI( "host:12345" ));
+
+    copy = copy;
+    BOOST_CHECK( copy != hostPort );
```
hernando: BOOST_CHECK( copy != zeq::URI( "host" ));

zeq/publisher.cpp
hernando: Maybe /proc/self/exe doesn't exist, for example is someone tries in FreeBSD (also a UNIX flavour), and it will silently fail.

zeq/publisher.h
```diff
@@ -10,6 +10,8 @@
 #include <zeq/api.h>
 #include <zeq/types.h>
 
+#include <memory>
+
 namespace zeq
 {
 namespace detail { class Publisher; }
```
hernando: remove

zeq/publisher.h
```diff
@@ -60,16 +115,19 @@ class Publisher
      * constructor uri.
      *
      * @return the publisher URI.
+     * @todo change signature to return zeq::URI, needs downstream project
+     *       adaptions. Also make zeq::URI( const servus::URI& from ) explicit.
      */
     ZEQ_API const servus::URI& getURI() const;
 
-    /** @return the port to which this publisher is bound. */
-    ZEQ_API uint16_t getPort() const;
```
hernando: Since this is a hard API change, people migrating code need to know what's the alternative. Document it in the Changelog or keep this function deprecated if possible.

zeq/publisher.h
```diff
@@ -17,17 +19,70 @@ namespace detail { class Publisher; }
...
+    ZEQ_API Publisher();
+
+    /**
+     * Create a publisher which announces the specified session.
+     *
+     * Constrains:
+     * - bound to all network interfaces
+     * - runs on a random port
+     * - announces itself on the _zeroeq_pub._tcp ZeroConf service as host:port
+     * - announces given session or ZEROEQ_SESSION from environment if set
```
hernando: Shouldn't the explicit session provided on the constructor override ZEROEQ_SESSION?

zeq/publisher.h
```diff
@@ -17,17 +19,70 @@ namespace detail { class Publisher; }
...
+    /**
+     * Create a publisher which runs on the specified URI.
+     *
+     * Constrains:
+     * - bound to the host and/or port from the given URI
+     * - announces itself on the _zeroeq_pub._tcp ZeroConf service as host:port
+     * - announces session \<username\> or ZEROEQ_SESSION from environment if set
+     *
+     * @param uri publishing URI in the format [scheme://][*|host|IP|IF][:port]
+     * @throw std::runtime_error if session is empty or socket setup fails
```
hernando: how can the session be empty? C&P bug?

zeq/publisher.h
```diff
@@ -17,17 +19,70 @@ namespace detail { class Publisher; }
...
+    ZEQ_API explicit Publisher( const URI& uri );
+
+    /**
+     * Create a publisher which runs on the specified URI and announces the
+     * specified session.
+     *
+     * Constrains:
+     * - bound to the host and/or port from the given URI
+     * - announces itself on the _zeroeq_pub._tcp ZeroConf service as host:port
+     * - announces given session or ZEROEQ_SESSION from environment if set
```
hernando: same as above

zeq/publisher.h
tribal-tec: It's redundant to getURI(), only tests were using it and no @version indicates any stable API here :) I'll document in the Changelog.

zeq/publisher.h
```diff
@@ -17,17 +19,70 @@ namespace detail { class Publisher; }
 /**
  * Serves and publishes events, consumed by Subscriber.
  *
+ * The session is tied to ZeroConf announcement and can be disabled with
+ * zeq::NO_ANNOUNCE_SESSION.
```
hernando: "passing zeq::NO_ANNOUNCE_SESSION as the session name." At least that's what I guessed.

zeq/publisher.h
tribal-tec: Both handlings can be valid or convenient. Changing the env variable is imo easier than the hardcoded one, but one could argue that a hardcoded session should never be altered. Not sure what to do now.

zeq/publisher.h
tribal-tec: the env variable can be empty?

zeq/publisher.h
eile: The hardcoded session might in fact not be hardcoded but come from an UI element -- in which case it should take precedence over the env variable.

In Eq I have: env, overwritten by globals, overwritten by instance.

tools/eventGenerator.cpp
```diff
@@ -195,7 +192,7 @@ void parseScript( const char* filename, Events& events )
 void printUsageAndExit( const char* name, const int code, bool full = false )
 {
     std::cerr << "Usage: " << name
-              << " [--schema,-s uri] [script]" << std::endl
+              << " [script]" << std::endl
```
eile: What is this? Prints:
    Usage: ./eventGenerator [script]
           ./eventGenerator --help

Why not add --help in the Usage?

zeq/detail/broker.h
```diff
@@ -1,30 +1,58 @@
...
+{
+    return buildZmqURI( uri.getScheme(), uri.getHost(), uri.getPort( ));
 }
 
-std::string buildZmqURI( const servus::URI& uri )
+std::string getUserName()
 {
-    return buildZmqURI( uri.getHost(), uri.getPort( ));
+    const char* user = getlogin();
+    return user ? user : "";
```
eile: "Unknown user"

zeq/types.h
```diff
@@ -50,8 +50,12 @@ static const uint32_t TIMEOUT_INDEFINITE = 0xffffffffu;
 
 using servus::make_uint128;
 
+static const std::string DEFAULT_SESSION = "__zeq";
+static const std::string NO_ANNOUNCE_SESSION = "__no_announce";
```
eile: NULL_SESSION

zeq/uri.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "detail/constants.h"
+#include "log.h"
+
+#include <cassert>
+
+namespace zeq
+{
+namespace
+{
+servus::URI createURI( const char* string )
```
eile: const std::string?

zeq/publisher.h
hernando: Shouldn't that default to \<username\> then? Just to throw as the last alternative.

zeq/publisher.h
hernando: In program options I always expect *command line option > config file option > environmental variable > defaults*, I don't know if this is a de facto rule, but following the same reasoning I prefer the constructor parameter to take precedence.

tools/eventGenerator.cpp
hernando: Do you find it confusing?
./eventGenerator [script] [--help] is also fine.

zeq/publisher.h
rdumusc: +1

zeq/detail/broker.h
```diff
@@ -1,30 +1,58 @@
...
 }
 
+std::string getDefaultSession()
+{
+    const char* session = getenv( ENV_SESSION.c_str( ));
+    return session ? session : getUserName();
```
hernando: Check also if strcmp(session, "") == 0?

tests/pubSub.cpp
```diff
@@ -94,12 +80,10 @@ BOOST_AUTO_TEST_CASE(publish_receive_zeroconf)
...
 
-    zeq::URI uri = test::buildPublisherURI();
-    zeq::Publisher publisher( uri );
-    zeq::Subscriber noSubscriber( uri );
-
-    uri.addQuery( "subscribeSelf", "true" );
-    zeq::Subscriber subscriber( uri );
+    zeq::Publisher publisher( test::buildUniqueSession( ));
+    zeq::Subscriber noSubscriber( publisher.getSession( ));
+    zeq::detail::Sender::getUUID() = servus::make_UUID(); // different machine
```
hernando: I was reading the documentation of Subscriber and I was wondering, "how did he test subscription in the same machine"?

zeq/publisher.h
```diff
@@ -10,24 +10,78 @@
...
  * Example: @include tests/publisher.cpp
  */
 class Publisher
 {
 public:
     /**
-     * Create a publisher on the given URI.
+     * Create a default publisher.
+     *
+     * Constrains:
```
hernando: Constrain is a verb, the noun is constraint.
But there are more postconditions than constraints, aren't they?

zeq/subscriber.h
```diff
@@ -19,6 +19,16 @@ namespace detail { class Subscriber; }
 /**
  * Subscribes to Publisher to receive events.
  *
+ * If the subscriber is in the same session as discovered publishers, it
+ * automatically subscribes to those publishers. Publishers from the same
+ * application instance are not considered though.
```
hernando: Is there a way to connect publishers and subscribers from the same application without having to use explicit URIs?

zeq/subscriber.h
```diff
@@ -27,27 +37,77 @@ class Subscriber : public Receiver
...
+     */
+    ZEQ_API explicit Subscriber( const URI& uri );
+
+    /**
+     * Create a subscriber which subscribes to publisher(s) the from given
+     * session.
+     *
+     * Constrains:
+     * - discovers publishers on _zeroeq_pub._tcp ZeroConf service
+     * - filters for given session or ZEROEQ_SESSION from environment if set
```
hernando: For symmetry to the publisher the session argument should take precedence over ZEROEQ_SESSION and an empty string should throw.

hernando:
Review missing @version tags

tribal-tec:
Do we want that for pre-major 1/non-stable APIs?

zeq/subscriber.h
tribal-tec: No.

hernando:
OK fine, actually I hadn't notice there aren't any version tags.

tribal-tec:
Any more comments? Any comments I missed?

eile:
+1 for me

rdumusc:
A lof of work, and it looks very good!
I only disagree with the subclassing of servus::uri. Making a class hierarchy introduces extra complexity and many opportunities for misusing the class through pointers for no good reason IMHO - when all that is needed is the createURI() helper function.
Read for instance: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#-cconcrete-concrete-types
Also the destructor of servus::URI is not virtual, and the zeq::URI class invariants are checked through asserts which I think is not sufficient because it will let production code behave incorrectly.
But of course I'm happy to hear other opinions here.

hernando:
What about using private inheritance and "using foo" declarations for functions to expose? That avoids the non-virtual destructor problem because zeq::URI can't be used as a servus::URI.

tribal-tec:
I like the idea of @hernando . It breaks unfortunately more downstream projects, because Publisher::getURI() can no longer return a servus::URI. So virtual dtor for servus::URI or API breakage?

rdumusc:
I think I have never used private inheritance and had to google to be sure of the consequences ;-)
But it makes the public functions of the base class private, so then you have to whitelist the parts which have to be public. I don't think this is what we want here, if only because it is very unusual. I'm sure wwe can do better. What about a simple zeq::URI class with the desired constructor and a toServusURI() method which creates and returns an appropriate servusURI object by value?

hernando:
Another option is to use aggregation instead of inheritance, so it doesn't even need to create the servus::URI by value. This solution fulfills all the requirements at the expense of being the most verbose one.

rdumusc:
sure, the servus::URI can also be a member of zeq::URI. I think this is preferable than starting a class hierarchy.

tribal-tec:
Private inheritance behaves almost exactly like aggregation/composition, so providing a toServusURI() is trivial and return-by-reference comes automatically. Verbosity as a bit less as only a couple of using declarations are necessary, in contrast to forward all calls. Private inheritance is not common, but I think in this case it serves quite well.

hernando:
I prefer private inheritance provided zeq::URI::toServusURI returns const& and you use that function in Publisher::getURI to avoid API breakage.

rdumusc:
I am still far from convinced. If private inheritance behaves like aggregation/composition why go for the most convoluted approach? Unless I'm mistaken the number of functions you would have to forward is close to zero, the only thing which seems to be used from zeq::URI is the constructor and a const servus::URI & getServusURI for passing downstream. But if you are 2 against 1, you win... :-)

tribal-tec:
Interesting. I see it the other way around. Why having a composition 'manually' (Pimpl then?) rather than using private inheritance. I needed to forward a couple of functions. In the end I'm not sure which one is better or more correct. Have a look now what you think. I just updated the commit.

eile:
Is there anything wrong with public inheritance and virtual dtors? Might not be the purest in OO terms, but results in the most sane usage in applications.

tribal-tec:
retest this please

rdumusc:
I think this new class with private inheritance should behave correctly.
@eile one of the problems with the public inheritance in this case is that you can't forbid users to call setSchema("") which defeats the puropse of having this zeq::URI in the first place.

hernando:
From the client's perspective I don't think using public inheritance with a virtual destructor is saner that the last commit. Both approaches are correct, despite not semantically equivalent. The proposed solution is more restrictive than using public inheritance, but that's actually good because setSchema can be hidden. I don't see what's the benefit of making the destructor in servus::URI virtual on the other hand.

eile:
@rdumusc: The purpose of the class is to parse URI( "host:port" ) in an zeq-specific way, and not as a path.

eile:
Afaik we never intended to hide setSchema.

rdumusc:
If I look at the implementation of the class, its main function is to prevent the scheme from being empty.
Simply parsing a "host:port" uri could be achieved directly in Servus with a factory function like:
static servus::URI fromHostAndPort( const std::string& uri );

rdumusc:
Or the same function in zeq to also set the schema if it is empty, but there would be no need for a class at all

tribal-tec:
From a user point-of-view, zeq::URI( "host:port" ) would not yield in what you would except, unless the user uses fromHostAndPort. I as a user would use different software then, because this API is not usable in my opinion.

rdumusc:
Yes, I am not sure if there could have been a cleaner way to achieve this, but at least we have a solution in the end... :-)

tribal-tec:
I'll merge in 1-2 hours unless I receive a veto :)

tests/publisher.cpp
```diff
@@ -67,17 +69,103 @@ BOOST_AUTO_TEST_CASE(multiple_publisher_on_same_host)
...
+                       "testsession" );
+    unsetenv(  "ZEROEQ_SESSION" );
+}
+
+BOOST_AUTO_TEST_CASE(empty_session)
+{
+    BOOST_CHECK_THROW( const zeq::Publisher publisher( "" ),
+                       std::runtime_error );
+}
+
```
hernando: What about a test for an empty ZEROEQ_SESSION. I don't know what's the expected behaviour in that case in the end.

tribal-tec: Right now empty ZEROEQ_SESSION results in filtering sessions for the current user. I was hoping for comments from others here as I'm not sure about this behavior as well.

rdumusc: Seems like a good default behaviour to me, that's how most users will want to use ZeroEQ!

Looking at this, I was wondering if we shouldn't use a small zeq::Session class, to gather the extraction of username + env variable and separate the responsibilities of deciding if a session is valid or not. I would be easy to unit test, and would make the Publisher constructor a bit more explicit. In the tests, it is hard to guess what Publisher( "" ) means. What do you think?

hernando: Defaulting to username if ZEROEQ_SESSION is empty is fine for me.

Regarding the session class I think it introduces a complication on the API unless it can be default constructed from std::string and cosnt char*. If you read the documentation I think it's pretty clear that Publisher( "" ) throws. Actually it's good that it throws because it doesn't leave the user wondering what would happen if accepted.
The only benefit I see about having the class is that you can query the default session before passing it to the Publisher if needed.

rdumusc: Agree with the throw, and the Session class constructor would take over this behaviour.
One can argue that it complicates the API; but only a little so, and only when using a session parameter. I see it as helping separate concerns and disambiguates better between the URI and Session constructors of Publisher, both based on strings. But to be clear it is just a suggestion and I am not blocking the PR :-)

tribal-tec:
Can we move this to a separate commit/PR if we decide to have this Session class? I would reaaaally like to get this merged!

rdumusc:
yes, repeating myself: I did not intend to block this PR

PR #110, opened 2015-11-19
Fix OS X build

eile:


eile:
hold this - more to come

eile:
Good to go.

PR #114, opened 2016-01-08
HTTP server spec

eile:
Formatted at https://github.com/eile/zeq/blob/master/doc/feature/httpServer.md

### DIFF
--- a/doc/feature/httpServer.md
+++ b/doc/feature/httpServer.md
@@ -1,0 +1,185 @@
+HTTP Server for ZeroEq Events
+============
+
+The http server implements a zeq::Receiver and Sender to serve ZeroBuf
+objects through a REST interface with JSON payload. It is the evolution
+of the RESTBridge.
+
+## Requirements
+
+* Translate Zerobufs to and from JSON
+* Accept PUT and POST requests and update subscribed Zerobuf in the same
+  way as a zeq::Subscriber
+* Accept GET requests and reply with the current state of registered
+  Zerobuf in the same way as a Publisher::publish().
+* Allow custom code execution while handling a GET request, e.g., to
+  render a new frame before serving a ImageJPEG request
+* Implement the same client REST API as the current RESTBridge
+
+## Dependency Changes
+
+* Remove: RESTBridge, cppnetlib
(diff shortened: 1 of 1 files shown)


eile:
@tribal-tec @bilgili @hernando @rdumusc FYI

doc/feature/httpServer.md
```diff
@@ -0,0 +1,144 @@
+HTTP Server for ZeroEq Events
+============
+
+The http server implements a zeq::Receiver and Sender to serve ZeroBuf
```
bilgili: The idea is nice but before implementing anything, I highly encourage "us" to do some research what is there already.
- The http server implementation is scary ( although it can be 30 lines: http://hintjens.com/blog:42 ). When, for instance https etc is needed we may need to look for other solutions.
- There are some libraries like http://mongrel2.org/ and/or https://github.com/fanout/zurl we may need to look as well.
- For the JSON ( as we have decided text over cable protocol ), we may check possible JSON libraries ( mongrel2 has it I think )

eile:
@bilgili:
1) See the zmq API doc for the http server part. Not more scary then the other zmq code we have for pub/sub, and the HTTP parsing is delegated to a library
2) I considered, but not documented them. done now.
3) We already have and use a JSON library (jsoncpp)

favreau:
1) RESTBridge was not meant to be ZeroEQ specific in its design. Removing this library forces the use of ZeroEQ to anyone willing to expose a REST interface in a C++ application. That said, I am always happy to remove dependencies, and AFAIK we have no plans to use anything else but ZeroEQ in the near. So I guess it's fine.
2) HTTPS/SSL should, IMHO, not be handled by the C++ application anyway. In the case of the Visualization Web Services, the authentication mechanism lies between the Rendering Resource Manager and the external world. The communication between the Rendering Resource Manager and the C++ applications can then safely use non-encrypted communication. HTTPS/SSL should only be considered if we want to expose C++ applications to the external world.

eile:
@favreau

1) addressed through a Servus::Serializable interface
2) Nice, issue 3 is decided then

favreau:
+1

PR #122, opened 2016-02-01
http::Server: Send error as body to be more informative

eile:


tribal-tec:
Not being an http expert though, but +1

PR #125, opened 2016-02-09
Made http urls and type names case insensitive

favreau:


zeq/http/server.cpp
```diff
@@ -225,6 +226,15 @@ class Server::Impl : public detail::Sender
...
             response.set_status( 400 );
     }
+
+private:
+    std::string _toLower( const std::string& value )
+    {
+        std::string str( value );
+        std::transform( str.begin(), str.end(), str.begin(), ::tolower );
+        return str;
+    }
```
eile: Can be a free function in an anonymous namespace.

zeq/http/server.cpp
```diff
@@ -45,7 +46,7 @@ class Server::Impl : public detail::Sender
 
     bool subscribe( servus::Serializable& serializable )
     {
-        const std::string& name = serializable.getTypeName();
+        const std::string name = _toLower( serializable.getTypeName( ));
```
eile: & can stay

bilgili:
Why do we need this change ? We are not following any standards but those are case insensitive in real world ?

favreau:
@bilgili There are lots of discussions around and the answer depends on the OS hosting the service. It seems to me that the majority of APIs around the web are case insensitive, and I personally tend to agree with this policy (http://stackoverflow.com/questions/7996919/should-url-be-case-sensitive).

bilgili:
fair enough

eile:
Test is missing.

zeq/http/server.cpp
```diff
@@ -14,9 +14,20 @@
...
 #include <httpxx/ResponseBuilder.hpp>
+#include <algorithm>
 
 namespace httpxx = ::http; // avoid confusion between httpxx and zeq::http
 
+namespace
+{
+std::string _toLower( const std::string& value )
+{
+    std::string str( value );
```
eile: rm this line, pass value by copy.

eile:
+1 otherwise

eile:
+1, rebase master missing.

PR #131, opened 2016-02-17
Specification for REST access to application API

favreau:


doc/feature/restAPI.md
```diff
@@ -0,0 +1,56 @@
...
+and invoked.
+
+## Requirements
+
+* Define how to invoke the application REST API
+* Implement the JSON representation of the REST API
+
+## API
+
+Invoking a GET resquest on the /api url returns a JSON representation of the
```
eile: typo resquest

doc/feature/restAPI.md
```diff
@@ -0,0 +1,56 @@
...
+schemas for all registered serializables.
```
eile: We previously discussed: ```GET /``` returns list of top-level namespaces, e.g., 'zerobuf', ```GET /zerobuf/render/``` would return the objects in zerobuf::render. ```GET /namespace/object/schema``` would return the schema for the object.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+- Query API
+   
+   Request: GET http://hostname:port/api
+   Response: { api: { "object1": "Schema1", "object2": "Schema2" } }
+
+## Implementation
+
+* Add a getJSONSchema public method to Servus::serializable
```
eile: I would just call it getSchema, the fact that it's in JSON format is a detail. Please document the C++ API above.

doc/feature/restAPI.md
favreau: I had something different in mind, and I it was more like what Django (the python implementation that we currently use for the Visualization Web Services) does. That is returning available URIs a as list when you do a GET on the /api uri of the server. What you are suggesting requires more processing on the client side, with a recursive walk in the tree of URIs. I would vote for a simple and single request for getting that list. But we can discuss again. This would simplify both the client and server side implementations.

eile: Can you make this an issue, discuss alternatives and propose a solution? I'm not against /api per se.

favreau: Sounds cool, on it!

favreau:
Ping

doc/feature/restAPI.md
```diff
@@ -0,0 +1,94 @@
...
+return the schema for the object itself.
+
+   Request: GET http://hostname:port/
+   Response: [ "namespace1", "namespace2", "..." ]
+
+   Request: GET http://hostname:port/namespace1/
+   Response: [ "object1", "object2", "..." ]
+
+   Request: GET http://hostname:port/namespace1/object1/schema
+   Response: { "object1" : { "attribute1": "string", "attribute2": "boolean" } }
```
tribal-tec: what's the C++ API changes/additions to enable this? Or done by default on register/subscribe?

doc/feature/restAPI.md
```diff
@@ -0,0 +1,94 @@
...
+* Add a getSchema public method to Servus::serializable
+* Make sure getSchema is added to code generator
+* Add the '/schema' url to every registered object, and make it return the
+  schema of the object in a JSON form.
+
+## Issues
+
+### API
+
+Several options are beging discussed regarding the implementation of the API:
```
tribal-tec: typo 'beging'

doc/feature/restAPI.md
```diff
@@ -0,0 +1,94 @@
...
+   Response: [ "namespace1", "namespace2", "..." ]
+
+   Request: GET http://hostname:port/namespace1/
+   Response: [ "object1", "object2", "..." ]
+
+   Request: GET http://hostname:port/namespace1/object1/schema
+   Response: { "object1" : { "attribute1": "string", "attribute2": "boolean" } }
+- Pros: Clean API, and a one-to-one match between a URI and the registered 
+  object.
+- Cons: Client needs to traverse the tree or URIs to get the list of registered
```
tribal-tec: If the specification defines that there is always a namespace after host:port, there is no ambiguity anymore.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,94 @@
...
+   Response: 
+   { 
+      { "namespace/foo": { foo: { "name": string } },
+      { "namespace/bar": { bar: { "x": float, "y": float, 'z": float } 
+   }
+- Pros: Simple client and server side implementations. No need for a recursive 
+  walk in the tree of URIs. Only one invocation is required to get the list of
+  all registered objects.
+- Cons: /api-docs becomes a reserved URI.
+
```
tribal-tec: We should already decide which of the options to implement, right? I guess you favor Option2, and it seems more prominent in the web world. I'd favor Option1 right now because of the clean semantics, but introspection is cumbersome, agreed. I'm not sure how we vote for that.

doc/feature/restAPI.md
favreau: I've added some indications in the Implementation section of the document. I am voting for the 'by default' option. This mechanism should be transparent for the user of the class.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,131 @@
...
+using the a GET operation on the URI of that object, suffixed by /schema.
+- Example:
+  - Get URIs to registered objects:
+    Request: GET http://hostname:port/registry
+    Response:
+    {
+      { "namespace/foo": "GET" },
+      { "namespace/foo": "PUT" },
+      { "namespace/bar": "PUT" }
+    }
```
eile: Doesn't format nicely in preview

eile: I think one object should have one entry, i.e, ```{ "namespace/foo": [ "GET", "PUT" ] },```

doc/feature/restAPI.md
```diff
@@ -0,0 +1,131 @@
...
+   Response: { "camera" : { "origin": "string", "lookat": "up" } }
+
+## Implementation
+
+- _registry_ and _schema_ are both keywords that correspond to GET requests.
+  They both need to be handled by the _processGet method of the
+  zeq::http::Server implementation.
+- This mechanism should be transparent to the user of the class, and should be
+  handled internally to the zeq::http::Server. The register/subscribe methods
+  must store the schemas together with the internal containers for registrations
```
eile: No, they can call getSchema() at request time.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,131 @@
...
+  PUT)
+- Add a getSchema public method to Servus::serializable
+- Make sure getSchema is implemented by to the code generator and returns the
+  content of the .fbs file.
+
+## Issues
+
+### API
+
+Several options are being discussed regarding the implementation of the API:
```
eile: Add _Resolution: Option x:_ $Reason.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,129 @@
...
+
+    // Register C++ objects
+    subscribers.push_back( _httpServer );
+    _httpServer.subscribe( camera, LUT, vrParams );
+    _httpServer.register( camera, LUT, vrParams, frame );
+  }
+```
+Query API:
+   * Request: ```GET http://hostname:port/zerobuf/render/camera/schema```
+   * Response: ```{ "camera" : { "origin": "string", "lookat": "up" } }```
```
tribal-tec: type of origin is more like zerobuf/render/vector3f. Please elaborate how nested ZeroBufs are exposed and handled in an example.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,129 @@
...
+  the contents of _registrations and _subscriptions containers in the form of
+  a JSON list a tuples defined by a URI and a list of REST operations (currently
+  GET and PUT).
+- This mechanism should be transparent to the user of the class, and should be
+  handled internally to the zeq::http::Server. The schema can be requested at
+  any time on invocation of the getSchema method on the registered object.
+
+## Issues
+### API
+Three options are being discussed regarding the implementation of the API:
```
tribal-tec: Issue is more a list of questions that need to be resolved. In this case we should document which of the three options was chosen and why.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,129 @@
...
+
+* Define how to invoke the application REST API
+* Implement the JSON representation of the REST API
+
+## API
+
+GET /registry returns the URIs for all registered objects, together with the
+supported REST operation (GET or PUT). Each objet can be queried for its schema
+using a GET operation on the URI of that object, suffixed by /schema.
+
```
tribal-tec: Document corner-cases: object/registry or namespace/schema are invalid, cause of 404 or something else.

doc/feature/restAPI.md
```diff
@@ -0,0 +1,129 @@
...
+    }
+```
+  - Get JSON schema for a specific object:
+    * Request: ```GET http://hostname:port/namespace/foo/schema```
+    * Response:```{ "foo": { "name": string, "age": integer } }```
+
+## Examples
+Create an HTTP server and register serializables:
+```
+  void setupHTTPServer( const int argc, char** argv )
```
tribal-tec: this code already exists due to prior spec. The new stuff here is 'only' the REST API.

tribal-tec:
Anything happening here?

favreau:
Was tired yesterday evening and messed up the PR when resync with master... Reopened a new PR.

PR #140, opened 2016-04-05
Added getSocketDescriptor() to http::server

rdumusc:


eile:
+1

zeq/http/server.cpp
```diff
@@ -310,6 +310,18 @@ const servus::URI& Server::getURI() const
     return _impl->uri.toServusURI();
 }
 
+SocketDescriptor Server::getSocketDescriptor() const
```
bilgili: Returning a low-level fd from the Server seems fishy to me.

rdumusc: Having access to the socket handle is very useful in conjunction with a QSocketNotifier. It handles the poll() as part of the QEventLoop and emits an activated() QSignal when the socket has new data available. This signal can be connected to a lamba which calls recieive(), so with a few lines of code you have a server which automatically receives without the need for a separate processing thread. I see no easy alternative to that mechanism.

bilgili: I totally understand it. But still looks fishy :)

eile: It's a commonly used pattern. You get an opaque handle to poll, but should not use it for something else. ZMQ provides it, X11 ConnectionNumber does, and many more.

bilgili: I did not disagree, but it feels like there should be extendable layer around it. A socket and poll classes. In C world it is more understandable but in C++ world these can be wrapped nicer.

PR #141, opened 2016-04-07
Remove deprecated constructors for using zeq::URI

bilgili:


tribal-tec:
Changelog update and version bump missing. +1 otherwise.
Merge should be done once release is out and downstream projects are ready as well.

tribal-tec:
Hold on, not complete yet:

1. zeq::Publisher::getURI() and zeq::http::Server::getURI() must return zeq::URI as well (see todo)
2. zeq::URI( const servus::URI& ) needs to be explicit (see todo)
3. zeq/uri.cpp:28 has TODO to be resolved

bilgili:
So, additionally all "todos" have to be resolved for this topic.

tribal-tec:
The mentioned todos are part of this topic, yes. Not sure if it's all todos in the project.

tests/uri.cpp
```diff
@@ -38,14 +38,6 @@ BOOST_AUTO_TEST_CASE(host_port)
     BOOST_CHECK_EQUAL( hostPort.getScheme(), "tcp" );
 }
 
-BOOST_AUTO_TEST_CASE(custom_schema_not_supported_yet)
```
tribal-tec: that test is still valid imo.

bilgili: // TODO: only for compatibility for hbp schemas, remove after downstreams		
// are fixed
I removed this to do from uri.cpp. What is in mind for this then ?

tribal-tec: The schema is left untouched, so actually only line 46 changes. The testname is then custom_schema. For the zeq::URI this is supported then, but not yet for the publisher/subscriber. Will do later then when needed. inproc could be interesting to have 'local' GUI only for instance.

PR #142, opened 2016-04-12
Added access control for direct communication

favreau:
Adding Access-Control-Allow-Origin to the response header allows direct communication between a browser and the c++ application.

### DIFF
--- a/tests/http/server.cpp
+++ b/tests/http/server.cpp
@@ -19,14 +19,22 @@ namespace
 const std::string jsonGet( "Not JSON, just want to see that the is data a-ok" );
 const std::string jsonPut( "See what my stepbrother jsonGet says" );
 
-const std::string error400( "HTTP/1.0 400 Bad Request\r\nContent-Length: 28\r\n\r\nHTTP/1.0"
-                            " 400 Bad Request\r\n\r\n" );
-const std::string error404( "HTTP/1.0 404 Not Found\r\nContent-Length: 26\r\n\r\nHTTP/1.0"
-                            " 404 Not Found\r\n\r\n" );
-const std::string error405( "HTTP/1.0 405 Method Not Allowed\r\nContent-Length: 35\r\n\r\nHTTP/1.0"
-                            " 405 Method Not Allowed\r\n\r\n" );
-const std::string error411( "HTTP/1.0 411 Length Required\r\nContent-Length: 32\r\n\r\nHTTP/1.0"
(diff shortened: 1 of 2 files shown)


eile:
From my limited http understanding, isn't this a site configuration option, i.e., should this not added by a proxy server or through some configuration option rather then allowing '*' to access the application?

favreau:
This is indeed managed by the proxy server in a deployed configuration. In which case, the security is managed between the end-client and the proxy server. This issue is about connecting a web browser directly to the C++ application. In that very case, no security is considered and allowing '*' seems fine to me. If we don't do that, we will always need a proxy between the end-client and of the C++ application.

favreau:
Further testing shows that PUT request are still being rejected... Investigating...

favreau:
PUT rejection was caused by "preflighted" requests. In 2 words, an OPTIONS request is sent to the server before the PUT to make sure that the update is accepted. In the current implementation, the OPTIONS method is not implemented is always return a 401 error. To be coherent with this PR, OPTIONS must be accepted so that PUT requests also are.

eile:
Please condense this into a spec

favreau:
The spec is there: https://www.w3.org/TR/cors, and the case that I am describing is in the introduction paragraph: '... the author needs to reply to a preflight request that uses the OPTIONS method and then needs to handle the actual request that uses the desired method (...) and give an appropriate response'.

favreau:
Retest this please

eile:
Right now this PR contains the bare metal code. I'm sure in three months @rdumusc will wonder what all these headers are good for. ;)

favreau:
Let's discuss. All is said in the specifications and there is not much I can do apart from repeating what's on the w3 website. CORS headers is a security mechanism implemented by all browsers, and they expect the HTTP servers to understand it. Since the ZeroEQ implements its own http server, it has to deal with those CORS headers. So we have two solutions there: 1) We implement the CORS headers in the ZeroEQ (good luck there!), 2) We allow every request and do not deal with CORS headers in ZeroEQ, and delegate that processing to an external proxy. What I implemented in this PR is solution 2. Let's discuss if it's not clear for everyone.

eile:
I'm fine with the implementation - but it's missing the context as to the why. All I'm asking for is some pointers in the code or a spec file.

favreau:
Ok, sorry, misunderstood you there. I am adding the w3 link in the code.

favreau:
Closed by mistake

PR #143, opened 2016-04-12
Rename zeq to ZeroEQ in the whole project

rdumusc:


CMake/zeq.cmake
```diff
@@ -1,15 +1,2 @@
...
-
-if(CI_BUILD_COMMIT)
-  set(ZEQ_REPO_TAG ${CI_BUILD_COMMIT})
-else()
-  set(ZEQ_REPO_TAG master)
-endif()
-set(ZEQ_FORCE_BUILD ON)
-set(ZEQ_SOURCE ${CMAKE_SOURCE_DIR})
+set(ZEROEQ_DEPENDS hbpvis REQUIRED Threads libzmq FlatBuffers Boost Servus)
+set(ZEROEQ_DEB_DEPENDS libboost-test-dev libzmq3-dev)
```
tribal-tec: move the second to the top-level CMakeLists before include(Common). The first line and this file can be removed. Would cause less conflicts on my end :)

rdumusc: Perfect, will do!

tribal-tec:
Good for me!

PR #145, opened 2016-04-13
Event as a servus::Serializable

bilgili:
It is not complete yet but has the idea. I am still debugging. FYI @eile.


zeq/event.cpp
```diff
@@ -49,9 +57,38 @@ flatbuffers::Parser& Event::getParser()
...
+bool Event::_fromBinary( const void *data, const size_t size )
+{
+    _impl->setData( ConstByteArray((uint8_t *)data, []( const void* ){}), size );
+    return true;
+}
+
+servus::Serializable::Data Event::_toBinary() const
+{
+    servus::Serializable::Data data;
+    data.ptr.reset( _impl->getData(), []( const void* ){});
```
bilgili: The servus::Data class uses a shared_ptr which is not really used correctly ( all uses are deleting the deleter which is at the end a raw ptr - check ZerBuf class _toBinary function as well ).

tests/connection/broker.cpp
```diff
@@ -33,8 +33,11 @@ class Subscriber
     void run()
     {
         zeq::Subscriber subscriber( test::buildURI( "127.0.0.1", *_publisher ));
-        BOOST_CHECK( subscriber.registerHandler( zeq::vocabulary::EVENT_ECHO,
-           std::bind( &Subscriber::onEchoEvent, this, std::placeholders::_1 )));
+        zeq::Event echoEvent( ::zeq::vocabulary::EVENT_ECHO,
+                              std::bind( &Subscriber::onEchoEvent,
+                                         this,
+                                         std::placeholders::_1 ));
```
eile: use lambda?

bilgili: Done

zeq/subscriber.cpp
```diff
@@ -190,47 +143,17 @@ class Subscriber::Impl
         zmq_msg_close( &msg );
 
         SerializableMap::const_iterator i = _serializables.find( type );
```
eile: error handling if i == _serializables.end();

zeq/types.h
```diff
@@ -44,8 +44,6 @@ class Subscriber;
 class URI;
 struct EventDescriptor;
 
-typedef std::shared_ptr< const uint8_t > ConstByteArray;
-
 typedef std::vector< EventDescriptor > EventDescriptors;
 typedef std::function< void( const Event& ) > EventFunc;
```
eile: This and _eventFuncs need removal

bilgili: EventFunc is valid.

eile: Ok, but _eventFuncs in Subscriber are not.

bilgili:
@eile : Are we renaming the zeroeq::Event ?

tests/connection/broker.cpp
```diff
@@ -32,10 +32,12 @@ class Subscriber
 
     void run()
     {
-        zeroeq::Subscriber subscriber( test::buildURI( "127.0.0.1",
-                                                       *_publisher ));
-        BOOST_CHECK( subscriber.registerHandler( zeroeq::vocabulary::EVENT_ECHO,
-           std::bind( &Subscriber::onEchoEvent, this, std::placeholders::_1 )));
+        zeroeq::Subscriber subscriber( test::buildURI( "127.0.0.1", *_publisher ));
```
eile: Same as for your code: Please adhere to 80c/line

zeroeq/event.h
```diff
@@ -9,33 +9,38 @@
...
 public:
     /**
      * Construct a new event of the given type
      *
      * @param type the desired event type
+     * @param func the desired event function
      * @sa vocabulary::registerEvent
      */
-    ZEROEQ_API explicit Event( const uint128_t& type );
+    ZEROEQ_API Event( const uint128_t& type, const EventFunc& func = EventFunc( ));
```
eile: Do we have a place where we don't use an eventfunc?

bilgili: unsubscribe( obj )

tests/connection/broker.cpp
bilgili: The difference is, I would not complain :)

zeroeq/event.cpp
```diff
@@ -9,19 +9,20 @@
...
 }
 
 Event::~Event()
 {
-    delete _impl;
+}
+
+std::string Event::getTypeName() const
+{
+    return _impl->type.getString();
```
eile: This is not the fully qualified, demangled class name. Also it's not clear to me how the JSON code gets its schema to do the serialization. Probably we should forbid REST/JSON for FlatBuffers.

zeroeq/event.h
eile: ? Then you use the existing instance.

eile:
> Are we renaming the zeroeq::Event ?

Probably. Best I could come up would be ```zerobuf::FlatBuffer```. FB should also become an optional dependency.

zeroeq/event.cpp
bilgili: I thought tests were handling that, so I will double check.

eile: The http server is tested with a fake Serializable.

bilgili:
I would keep the Event name and move to FBEvent ?

bilgili:
- "FB should also become an optional dependency." 

Can you please open this a bit ? There are many place that refers to FB in the code ( CMake uses, C++ uses ). We want the FB related part as a separate library ?

eile:
> I would keep the Event name and move to FBEvent ?

I won't argue, but at this point it is an object (serializable) not an event anymore. Events will come back eventually, I had a first discussion with @chevtche last week but we are far from "done" on this feature.

eile:
> optional dependency

I was thinking 'common_find_package(FlatBuffers)' and if/ifdef ZEROEQ_USE_FLATBUFFERS in the appropriate places (probably just the CMakeLists to include the event class if fb available).

bilgili:
Then, for instance Monsteer is depending on ZeroEq, which means in CMake of Monsteer, FLATBUFFERS_FOUND is used ? Or, still there is a way to use ZEROEQ_FLATBUFFERS variables over the place ?

eile:
On 20. Apr 2016, at 17:32, Ahmet Bilgili <notifications@github.com> wrote:

> Then, for instance Monsteer is depending on ZeroEq, which means in CMake of Monsteer, FLATBUFFERS_FOUND is used ? Or, still there is a way to use ZEROEQ_FLATBUFFERS variables over the place ?

Monsteer has (and will still need) a "common_find_package(FlatBuffers REQUIRED)”, so no need for conditionals.


Cheers,

Stefan.

tests/broker.h
```diff
@@ -42,16 +42,16 @@ zeroeq::URI buildURI( const std::string& hostname, const zeroeq::Publisher& to )
...
     BOOST_CHECK( event.getType() == zeroeq::vocabulary::EVENT_ECHO );
     const std::string message = zeroeq::vocabulary::deserializeEcho( event );
     BOOST_CHECK_EQUAL( echoMessage, message );
 }
 
-void onExitEvent( const zeroeq::Event& event )
+void onExitEvent( const zeroeq::FBEvent& event )
 {
-    BOOST_CHECK_EQUAL( event.getType(), zeroeq::vocabulary::EVENT_EXIT );
+    BOOST_CHECK_EQUAL( event.getType(), zeroeq::vocabulary::EVENT_ECHO );
```
hernando: Why renamed?

bilgili: EXIT event is removed

hernando: @chevtche closed #153 without merging. On the other hand, if it's removed, then you have to remove the function.

bilgili: Which function ?

bilgili: I got it :)

hernando:
Maybe I'm wrong because I'm late to this discussion, but to me it seems that implementing #152 and then adapting the FB code to use the new API will require less work and produce a better result than adapting the FB to use Serializable. The only thing I'm not sure is whether the current Subscriber::registerHandler would conflict with the proposal from #152.

bilgili:
I started earlier with the idea and I have already moved everything in Monsteer towards it. So, lets merge these first and then lets see if we still need #152.

hernando:
#152 will still be needed.

bilgili:
If we have a empty serializable event made of "type" I am not sure you need it. I believe we have cleaned the subscriber & publisher API for this reason.

hernando:
I'm not sure I understand what you mean.

bilgili:
retest this please

bilgili:
retest this please

bilgili:
ping

doc/Changelog.md
```diff
@@ -2,6 +2,13 @@
 
 # git master
 
+* [145](https://github.com/HBPVIS/ZeroEQ/pull/145):
+  * The zeroeq::Event class is renamed to zeroeq::FBEvent
+  * zeroeq::FBEvent is derived from the servus::Serializable
```
eile: rm 'the'

tests/CMakeLists.txt
```diff
@@ -6,18 +6,36 @@ if(NOT BOOST_FOUND)
...
+
+if(NOT FLATBUFFERS_FOUND)
+  list(APPEND EXCLUDE_FROM_TESTS connection/broker.cpp
+                                 hbp/serialization.cpp
+                                 newEvent.cpp
+                                 perf.cpp
+                                 publisher.cpp
+                                 pubSub.cpp
+                                 receiver.cpp
+                                 subscriber.cpp)
```
eile: Oh no... we should not loose these tests with FB. They should be run at least with a minimal test object (similar to Servus/tests/serializable.cpp) and optional extended tests using the FB objects

zeroeq/subscriber.cpp
```diff
@@ -191,47 +143,20 @@ class Subscriber::Impl
...
-                const size_t size = zmq_msg_size( &msg );
-                ConstByteArray data( new uint8_t[size],
-                                     std::default_delete< uint8_t[] >( ));
-                memcpy( (void*)data.get(), zmq_msg_data( &msg ), size );
-                event.setData( data, size );
-                assert( event.getSize() == size );
-                zmq_msg_close( &msg );
-            }
+        if( i == _serializables.cend( ))
+            ZEROEQTHROW( std::runtime_error( "Cannot process type " + type.getString( )));
```
eile: Received event for non-subscribed type

zeroeq/detail/fbevent.h
```diff
@@ -45,10 +42,14 @@ class Event
...
-    void setData( const ConstByteArray& data_, const size_t size_ )
+    void setData( const uint8_t* data_, const size_t size_ )
     {
         parser.builder_.Clear();
-        data = data_;
+        // Since we are working on the same copy with new servus::Serializable
+        // based Flat Buffer events ( not memory per event ),
+        // we are copying the memory
+        data.reset( new uint8_t[ size_ ] );
+        memcpy( data.get(), data_, size_ );
```
eile: This copy is not needed. The contract is that the memory is valid only in the updated func, which should be (is not right now) called while the zmq msg is still open.

zeroeq/detail/fbevent.h
```diff
@@ -41,13 +38,16 @@ class Event
...
         return parser.builder_.GetBufferPointer();
     }
 
-    void setData( const ConstByteArray& data_, const size_t size_ )
+    void setData( const uint8_t* data_, const size_t size_ )
     {
         parser.builder_.Clear();
+        // Since we are working on the same copy with new servus::Serializable
+        // based Flat Buffer events ( not memory per event ),
+        // we are copying the memory
```
eile: rm

zeroeq/detail/fbevent.h
```diff
@@ -58,12 +58,14 @@ class Event
     flatbuffers::Parser parser;
 
     /** setData() uses this instead of fbb during deserialization */
-    ConstByteArray data;
+    const uint8_t* data;
+
     size_t size;
```
eile: const

zeroeq/fbevent.cpp
```diff
@@ -0,0 +1,84 @@
...
+        registerDeserializedCallback( [this](){ _impl->func( *this ); } );
+}
+
+FBEvent::~FBEvent()
+{
+}
+
+std::string FBEvent::getTypeName() const
+{
+    return _impl->type.getString();
```
eile: throw

zeroeq/subscriber.cpp
```diff
@@ -178,59 +130,31 @@ class Subscriber::Impl
...
+
+       servus::Serializable* serializable = i->second;
+       if( payload )
+       {
+           zmq_msg_init( &msg );
+           zmq_msg_recv( &msg, socket.socket, 0 );
+           serializable->fromBinary( zmq_msg_data( &msg ),
+                                     zmq_msg_size( &msg ));
+           zmq_msg_close( &msg );
+       }
```
eile: else
   notifyUpdated();

zeroeq/vocabulary.cpp
```diff
@@ -5,59 +5,56 @@
...
+     flatbuffers::Parser parser;
+     if( !parser.Parse( schema.c_str( )))
+        ZEROEQTHROW( std::runtime_error( parser.error_ ));
+
+     std::string json;
+     flatbuffers::GeneratorOptions opts;
+     opts.base64_byte_array = true;
+     opts.strict_json = true;
+     GenerateText( parser, event.getData(), opts, &json );
+     return json;
```
eile: rm

zeroeq/vocabulary.h
```diff
@@ -27,97 +27,32 @@ namespace vocabulary
...
+ * Deserialize a SCHEMA to a json
+ * @param event is the event to be deserialized
+ * @param schema is the schema for json
+ * @return json
  */
-ZEROEQ_API std::string deserializeJSON( const Event& event );
-//@}
-
+std::string deserializeJSON( const FBEvent& event,
+                             const std::string& schema );
```
eile: rm

zeroeq/fbevent.cpp
bilgili: This is used by the serializable. TypeName is converted to typeidentifier in servus::serializable

uint128_t Serializable::getTypeIdentifier() const
{
    return make_uint128( getTypeName( ));
}

zeroeq/subscriber.cpp
bilgili: @eile opened an issue for this #158. Will be handled after some discussions. Current fromBinary() function is implicitly calling notification. It wont be called for events without a payload.

zeroeq/subscriber.cpp
```diff
@@ -178,59 +130,31 @@ class Subscriber::Impl
...
-            if( payload )
-            {
-                zmq_msg_init( &msg );
-                zmq_msg_recv( &msg, socket.socket, 0 );
-                serializable->fromBinary( zmq_msg_data( &msg ),
-                                          zmq_msg_size( &msg ));
-                zmq_msg_close( &msg );
-            }
-        }
+       zmq_msg_t msg;
```
rdumusc: the whole block has been shifted, now it has 3 spaces

zeroeq/fbevent.h
```diff
@@ -0,0 +1,73 @@
...
+     * @param type the desired event type
+     * @param func the desired event function
+     * @sa vocabulary::registerEvent
+     */
+    ZEROEQ_API FBEvent( const uint128_t& type, const EventFunc& func );
+
+    /** Move ctor @internal */
+    ZEROEQ_API FBEvent( FBEvent&& rhs );
+
+    /** @return the fully qualified, demangled class name. */
```
rdumusc: Docstring should mention that it is unavailable and throws an exception

rdumusc:
I'm giving a +1 since this has already been extensively reviewed, but I don't understand every details.

PR #152, opened 2016-04-20
Event-based communication spec

eile:


eile:
@chevtche @rdumusc @bilgili @tribal-tec FYI - Just a write up of previous discussions, but overall does seem quite straight-forward.

doc/feature/Events.md
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    void onExit( const void*, size_t )
+    {
+        ::exit( EXIT_SUCCESS );
+    }
+
+    void setupZeroEQ()
+    {
+        _httpServer.subscribe( zerobuf::data::OpenEvent::ZEROBUF_TYPE(),
```
hernando: ZEROBUF_TYPE() doesn't exist yet, does it? If that's the case, make it explicit.

doc/feature/Events.md
```diff
@@ -0,0 +1,76 @@
...
+    {
+        // as Subscriber::subscribe(), except HTTPEventFunc returns
+        // status code (200, 404, etc)
+    };
+    }
+
+## Examples
+    int onOpen( const void* data, const size_t size )
+    {
+        zerobuf::data::ConstOpenEventPtr open =
```
hernando: auto instead of zerobuf::data::ConstOpenEventPtr?

hernando:
Looks pretty reasonable, +1.

doc/feature/Events.md
eile: It does not exist yet anymore. ;)

Used to be there, was probably cleaned up because not needed, but can be added easily.

doc/feature/Events.md
eile: sure

eile: Reason I made it explicit here is to highlight the "const"

PR #159, opened 2016-05-22
Fix MSVC build

tribal-tec:


tests/CMakeLists.txt
```diff
@@ -28,7 +28,10 @@ if(NOT HTTPXX_FOUND)
 endif()
 
 if(NOT FLATBUFFERS_FOUND)
-  list(APPEND EXCLUDE_FROM_TESTS hbp/serialization.cpp
+  list(APPEND EXCLUDE_FROM_TESTS connection/broker.cpp
+                                 hbp/serialization.cpp
```
bilgili: We did not excluded those tests on purpose. Returning a SerializablePtr was done because of it. If they are not compiling with some other configuration we can fix maybe in another way.

tribal-tec:
Updated

tribal-tec:
Retest this please

tribal-tec:
Ping

PR #164, opened 2016-05-25
Fix compilation issues with flatbuffer events and cmake ( add_custom_…

bilgili:
…command is not triggered multiple times )


rdumusc:
+1

hernando:
Without looking at the code I guess the problem is that the macro/function is not properly creating a new target for each invocation.

bilgili:
It is properly creating it. The problem seems cmake related.

PR #165, opened 2016-05-25
Fix compilation issues with fb events (order independent file generat…

bilgili:
…ion)


rdumusc:
+1 hopefully this should fix it once and for all :-)

bilgili:
@hernando : If you use the list of files in the function for generating headers and use set_property(SOURCE ... APPEND PROPERTY OBJECT_DEPENDS output ) it works. But this is the simplest solution ( now knowing the fate of FB after Monsteer is moved to ZB )

PR #166, opened 2016-05-26
Implement event-based communication as per #152

tribal-tec:


hernando:
FYI @ptoharia

zeroeq/http/server.h
```diff
@@ -114,9 +114,35 @@ class Server : public zeroeq::Receiver
      */
     ZEROEQ_API bool subscribe( servus::Serializable& object );
 
+    /**
+     * Subscribe an event to receive HTTP PUT requests.
+     *
+     * Every receival of the event will call the registered callback function.
```
hernando: typo receival

zeroeq/http/server.h
```diff
@@ -114,9 +114,35 @@ class Server : public zeroeq::Receiver
...
+     * @param event the event name to receive PUT requests for during receive()
+     * @param func the callback function for serving the PUT request
+     * @return true if subscription was successful, false otherwise
+     */
+    ZEROEQ_API bool subscribe( const std::string& event, const PUTFunc& func );
+
+    /**
+     * Subscribe an event to receive HTTP PUT requests with payload.
+     *
+     * Every receival of the event will call the registered callback function.
```
hernando: same typo

zeroeq/http/server.h
tribal-tec: google says it's correct?!

hernando: Indeed it is... it looked wrong to me in the beginning.

tests/http/server.cpp
```diff
@@ -212,18 +212,30 @@ BOOST_AUTO_TEST_CASE(registration)
...
+
     BOOST_CHECK( server.subscribe( foo ));
     BOOST_CHECK( !server.subscribe( foo ));
     BOOST_CHECK( server.unsubscribe( foo ));
     BOOST_CHECK( !server.unsubscribe( foo ));
 
+    BOOST_CHECK( server.subscribe( "foo",
+                                   []( const std::string& ) { return true; }));
+    BOOST_CHECK( !server.subscribe( "foo",
+                                    []( const std::string& ) { return true; }));
```
hernando: Late to change these names, but since the http server does not implement pub/sub communication, and I always have to look into the documentation to find out which callback setter is for PUT and which one is for GET, I have the impression that these functions should have been named differently.

PR #168, opened 2016-05-30
Fix MSVC build

tribal-tec:
Find why here: http://stackoverflow.com/questions/27750343/function-overload-using-lambda-function-signature


eile:
+1

tribal-tec:
retest this please

tribal-tec:
retest this please

PR #169, opened 2016-05-31
Fix #157

### LINKED ISSUE #157: http server may hang in receive() when remote connection is closed abruptly
The application deadlocks at ::zmq::send() at lines 195-197 of server.cpp trying to send a return code 200:
```
if( ::zmq_send( socket, id, idSize, ZMQ_SNDMORE ) != idSize ||
        ::zmq_send( socket, rep.c_str(), rep.length(), more ) !=
        int( rep.length( )))
{
ZEROEQWARN << "Could not send HTTP response header: "
                          << zmq_strerror( zmq_errno( )) << std::endl;
return;
}
```

tribal-tec:


eile:
Uhm, that looks potentially dangerous if msgSize > ZMQ_SEND_BUFFER_SIZE

tribal-tec:
Updated

zeroeq/http/server.cpp
```diff
@@ -213,25 +213,51 @@ class Server::Impl : public detail::Sender
...
+    {
+        while( ::zmq_send( socket, data, length,
+                           flags | ZMQ_NOBLOCK ) != int( length ))
+        {
+            // could be disconnect, send buffer full, ...
+            if( zmq_errno() == EAGAIN )
+            {
+                detail::Socket pollItem;
+                pollItem.socket = socket;
+                pollItem.events = ZMQ_POLLERR;
```
eile: | POLLOUT?

eile:
Test for overflowing send buffer?

+1 otherwise.

tribal-tec:
Tried the overflow test, no luck yet...

zeroeq/http/server.cpp
```diff
@@ -213,25 +213,50 @@ class Server::Impl : public detail::Sender
...
+            // could be disconnect, send buffer full, ...
+            if( zmq_errno() == EAGAIN )
+            {
+                detail::Socket pollItem;
+                pollItem.socket = socket;
+                pollItem.events = ZMQ_POLLERR | ZMQ_POLLOUT;
+                if( ::zmq_poll( &pollItem, 1, timeout == TIMEOUT_INDEFINITE
+                                              ? -1 : timeout ) > 0 )
+                {
+                    // client still alive, send again
```
eile: if( socket.revents & ZMQ_POLLOUT ) return true;
// else POLLERR
return false;

?

tribal-tec:
POLLOUT won't work on OSX. Remove?

zeroeq/http/server.cpp
tribal-tec: No. Seems on OSX the socket is not cleaned/closed like on Unix...

PR #171, opened 2016-06-07
http server index spec

eile:
Pretty version: https://github.com/eile/ZeroEQ/blob/master/doc/feature/index.md

### DIFF
--- a/doc/feature/index.md
+++ b/doc/feature/index.md
@@ -1,0 +1,59 @@
+Landing page for http::Server
+============
+
+This specification outlines the API to add a custom landing page to the
+zeroeq::http::Server. Currently
+[Tide](https://github.com/BlueBrain/Tide/blob/master/tide/master/rest/RestInterface.cpp)
+uses a hack to achieve static html pages.
+
+## Requirements
+
+* Custom, runtime-changeable index.html landing page
+* Default landing page displays a pretty-formatted version of /registry
+* Allow dynamic pages
+
+## Dependencies
+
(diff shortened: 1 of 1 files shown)


rdumusc:
This is a reasonable default for most use cases, and it's clear and simple to use. I think it can be implemented as is, without conflicting with the following remarks:

We may not want to limit the feature to the case of a static index page. With the current situation,  dynamic pages can also be provided (for any arbitrary url) with just a minor variation of the StaticContent class in Tide. However, the semantics are ugly and users have to "hack" servus::Serializable::_toJson() to achieve this result. Maybe zeroeq could provide some base class to make things more clean:
class HtmlPage
{
virtual std::string getHtml() const = 0;
};
with
class Server
{
    void register_( const HtmlPage& page, const std::string& url );
};

eile:
+1 on the HtmlPage.

Not sure about the generalization to URLs - it feels like feature creep towards making http::Server like apache & co. Opinions?

rdumusc:
+1 for your change. I agree that we should code only what is required and nothing more. We can start with (static) homepage as you suggest.

As for the general question, there is a risk of feature creep, but I am afraid that these extra developments will be needed at some point. For instance, Tide could benefit from a more advanced web UI. It would complement the touch interface and replace the complicated master application & VNC setup. This implies delivering images and javascript in addition to html. In the absence of a proper API, this will happen through more "hacking" of StaticContent : public servus::Serializable. or switching to a different network library.

eile:
@rdumusc Thanks, noted as issue in spec.

PR #179, opened 2016-09-07
Fix http server issues with "Peer-Address" messages on Ubuntu 16.04

rdumusc:
These unexpected messages caused Tide's RestServer to block the main
thread after each http request in zmq_recv() at line 144 of
zeroeq/http/server.cpp.

The exact combination of operations in this patch is necessary to ensure
that none of the following unit tests block indefinitely:

- zeroeq/http/server - issue157
- tide/cpp/core/RestServerTests - testServerReturnsSimpleContent

and that normal receive operations (as tested with Tide's REST interface)
still work as expected (no messages skipped due to zmq_errno() == EAGAIN)

### DIFF
--- a/zeroeq/http/server.cpp
+++ b/zeroeq/http/server.cpp
@@ -132,21 +132,23 @@ public:
 
     void process( const uint32_t timeout )
     {
-        // Read request and body
+        // Read http request
         httpxx::BufferedRequest request;
-        std::string body;
         uint8_t id[256];
         int idSize = 0;
+        int flags = 0;
         while( !request.complete( ))
         {
             // id of client (used for reply)
-            idSize = ::zmq_recv( socket, id, sizeof( id ), 0 );
+            idSize = ::zmq_recv( socket, id, sizeof( id ), flags );
             if( idSize <= 0 )
             {
-                ZEROEQWARN << "HTTP server receive failed: "
-                           << zmq_strerror( zmq_errno( )) << std::endl;
+                if( flags == 0 || zmq_errno() != EAGAIN )
+                    ZEROEQWARN << "HTTP server receive failed: "
+                               << zmq_strerror( zmq_errno( )) << std::endl;
                 return;
             }
+            flags = 0;
 
             // msg body
             zmq_msg_t msg;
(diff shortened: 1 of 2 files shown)


zeroeq/http/server.cpp
```diff
@@ -139,8 +139,9 @@ class Server::Impl : public detail::Sender
         int idSize = 0;
         while( !request.complete( ))
         {
+            int flags = 0;
             // id of client (used for reply)
-            idSize = ::zmq_recv( socket, id, sizeof( id ), 0 );
+            idSize = ::zmq_recv( socket, id, sizeof( id ), flags );
```
eile: Is this still needed or just a debugging artefact?

eile: Yes, just found it below.

eile: Actually no, the assignment below has no side effect due to the scope of flags?

zeroeq/http/server.cpp
```diff
@@ -157,10 +158,21 @@ class Server::Impl : public detail::Sender
...
-                return;
+                // Filter-out empty "Peer-Address" messages interleaved in the
+                // communication on Ubuntu 16.04 / libzmq 4.1.4
+                if( data && std::string( data ) == "Peer-Address" )
+                {
+                    zmq_msg_close( &msg );
+                    flags = ZMQ_NOBLOCK;
+                    continue;
+                }
+                else
```
eile: else after continue not needed.

eile:
Can you merge the #157 test into this PR?

eile:
+1 otherwise

eile:
@favreau had also seen the Peer-Address msg, iirc.

zeroeq/http/server.cpp
rdumusc: You're right, this assignment has no effect, yet I was sure it had solved the last of my problems when I did it. I'll remove it and test everything again on Friday.

zeroeq/http/server.cpp
rdumusc: will do

rdumusc:
retest this please

rdumusc:
Updated with a fix that I think is cleaner and has been tested on all platforms. I don't know how this could be specifically unit tested, however.

eile:
Gah, this is the kind of $#@! I wanted to avoid with zmq - Collage has similarly nice code. At least we're one level higher now at the http protocol (as opposed to tcp).

PR #181, opened 2016-10-11
Specification for registry and schema API using REST

tribal-tec:


doc/feature/restAPI.md
```diff
@@ -0,0 +1,71 @@
...
+* _registry_ and _schema_ are both keywords that correspond to GET requests.
+  They both need to be handled by the _processGet method of the
+  zeroeq::http::Server implementation.
+* Add a getRegistry() method to the zeroeq::http::Server implementation that
+  returns the contents of _registrations_ and _subscriptions_ in the form of a
+  JSON list a tuples defined by a URI and a list of REST operations (currently
+  GET and PUT).
+* This mechanism should be transparent to the user of the class, and should be
+  handled internally to the zeroeq::http::Server. The schema can be requested at
+  any time on invocation of the getSchem()a method on the registered object.
```
eile: Typo getSchem()a

doc/feature/restAPI.md
```diff
@@ -0,0 +1,71 @@
...
+
+* Define how to invoke the application REST API
+* Implement the JSON representation of the REST API
+
+## API
+
+* _GET /registry_ returns the URIs for all registered objects, together with the
+  supported REST operation (GET or PUT).
+* Each object can be queried for its schema using a GET operation on the URI of
+  that object, suffixed by /schema.
```
hernando: Would it make sense to use a name that is an invalid C++ identifier, for example /.schema?

doc/feature/restAPI.md
```diff
@@ -0,0 +1,71 @@
...
+                "minItems": 16,
+                "maxItems": 16,
+                "default": [ 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1 ]
+            }
+        },
+        "required": ["matrix"],
+        "additionalProperties": false
+    }
+
+```
```
hernando: Examples are good, but specification by example is not sufficient, this is far from being an specification at all. If I had to give another example with different properties I'd no clue about how to do it. For example, if I have a property which is a string, the type is "string" or "str"?
Just think about the following, could you write unit tests for validating the JSON schema just with this?

Questions that I have:
- What is the title providing compared to the name?
- What other types are possible
- What times of properties can you have?
- Why do you use number and not integer or float for the element type?
- What makes array special in its definition?
- What is additionalProperties used for?

Maybe I'm missing that this schema is already specified somewhere else. If that's the case, please add the link.

doc/feature/restAPI.md
eile: Imo this would make an awkward REST API. You can still get a /myns/Schema/schema.

doc/feature/restAPI.md
tribal-tec: There is not much change compared to the PR from April, so I'm a bit surprised the schema format matters now...
As @eile pointed out, this reflects JSON schema (http://json-schema.org/), but in the scope for this spec the schema format is irrelevant. It's a contract between concrete servers and applications what format/language they want to use for describing their exposed objects.
The important part here is that a 'schema' per object exists and is served by the HTTP server. A remark here: it's not specified if 'schema' is mandatory for each registered object or not. Will put in the issues section along with more API changes that I saw during implementation.

hernando: If the schema is really irrelevant, then I would remove it completely from the example because it confuses people that tends to pay attention to details like me.

tribal-tec:
Updated. I kept the schema nonetheless to illustrate a real-life example, and mentioned than JSON schema is recommended.

eile:
+1

PR #182, opened 2016-10-19
Implement HTTP server REST API

tribal-tec:
* added support for /schema and /registry endpoints
* added ZeroEQHTTP target, moved from ZeroEQ library
* rename subscribe() and register_() to handlePUT/GET accordingly
* group RFCs in doxygen


zeroeq/http/server.cpp
```diff
@@ -26,6 +32,30 @@ std::string _toLower( std::string value )
     return value;
 }
 
+// http://stackoverflow.com/questions/5343190/how-do-i-replace-all-instances-of-a-string-with-another-string
```
eile: Tip: you can remove the ```/how-....``` part

zeroeq/http/server.cpp
```diff
@@ -26,6 +32,30 @@ std::string _toLower( std::string value )
...
+std::string _replaceAll( std::string subject, const std::string& search,
+                         const std::string& replace )
+{
+    size_t pos = 0;
+    while( (pos = subject.find( search, pos )) != std::string::npos )
+    {
+         subject.replace( pos, search.length(), replace );
+         pos += replace.length();
+    }
```
eile: can also be written as a for loop.

zeroeq/http/server.cpp
```diff
@@ -285,48 +328,77 @@ class Server::Impl : public detail::Sender
 protected:
     typedef std::map< std::string, PUTPayloadFunc > PUTFuncMap;
     typedef std::map< std::string, GETFunc > GETFuncMap;
-    PUTFuncMap _subscriptions;
-    GETFuncMap _registrations;
+    typedef std::map< std::string, std::string > SchemaMap;
+    PUTFuncMap _PUT;
+    GETFuncMap _GET;
```
eile: why not lower-case variable names?

tests/http/server.cpp
```diff
@@ -505,3 +522,109 @@ BOOST_AUTO_TEST_CASE(urlcasesensitivity)
...
+
+    bool running = true;
+    std::thread thread( [&]() { while( running ) server.receive( 100 ); });
+
+    Client client( server.getURI( ));
+    client.test( "GET /bla/bar/schema HTTP/1.0\r\n\r\n", error404, __LINE__ );
+
+    running = false;
+    thread.join();
+}
```
eile: Missing tests:
* Object named Schema: 'GET /foo/schema', 'GET /foo/schema/schema'
* Object named Registry: throw when registering without namespace, 'GET /foo/registry', 'GET /foo/registry/schema'

zeroeq/http/server.cpp
```diff
@@ -68,58 +98,71 @@ class Server::Impl : public detail::Sender
...
     }
 
-    bool unregister( const std::string& event )
+    bool removeGET( std::string event )
     {
-        return _registrations.erase( _toLower( event )) != 0;
+        event = _toLower( _replaceAll( event, "/", "::" ));
+        if( _put.count( event ) == 0 )
+            _schemas.erase( event );
+        return _get.erase(  event ) != 0;
```
rdumusc: style opinion # 2: I don't feel that the != 0 adds anything

zeroeq/http/server.cpp
```diff
@@ -68,58 +98,71 @@ class Server::Impl : public detail::Sender
...
     }
 
-    bool register_( std::string event, const GETFunc& func )
+    bool handleGET( std::string event, const std::string& schema,
+                    const GETFunc& func )
     {
-        event = _toLower( event );
-        if( _registrations.count( event ) != 0 )
+        event = _toLower( _replaceAll( event, "/", "::" ));
+        if( _get.count( event ) != 0 )
```
rdumusc: general style opinion: I find if( !_put.count( event )) more readable

tests/http/server.cpp
eile: all GETs mentioned should return meaningful values, i.e., should not fail. I don't see in the spec that using an object named schema is forbidden, and I don't see a reason why it should be. idem for registry, except in the global namespace.

zeroeq/http/server.cpp
eile: Because it is one. Precondition is before while, loop statement is at the end of the while scope. Not important, though.

zeroeq/http/server.cpp
```diff
@@ -285,48 +328,77 @@ class Server::Impl : public detail::Sender
...
+
+    std::string _returnRegistry() const
+    {
+        Json::Value body( Json::objectValue );
+        for( const auto& i : _get )
         {
-            const size_t pos = name.find( '/' );
-            if( pos == std::string::npos )
-                return name;
+            const std::string key = _replaceAll( i.first, "::", "/" );
```
rdumusc: wouldn't it be more logical to store the keys with '/' directly? Only the servus::Serializable need a conversion at insertion time, no?

zeroeq/http/server.cpp
tribal-tec: Not my implementation, hence the reference of origin. Worked at first shot, didn't take a second glance to make it even better.

tests/http/server.cpp
tribal-tec: object named schema is tested, registry can be implemented, although I miss the use-case. global namespace registry is needed though, probably we're talking about different things here.

zeroeq/http/server.cpp
hernando: I don't know if it's relevant, but MSVC used to whine about a potential performance bug in the implicit int to bool conversion if you do that.

zeroeq/http/server.cpp
tribal-tec: To me we could use '/' everywhere, also from the ZeroBuf code generator. I'm trying to be as less invasive as possible and not changing things w/o discussing them.

zeroeq/http/server.cpp
```diff
@@ -26,6 +32,30 @@ std::string _toLower( std::string value )
...
+
+const std::string REQUEST_REGISTRY = "registry";
+const std::string REQUEST_SCHEMA = "schema";
+
+bool _endsWithSchema( const std::string& uri )
+{
+    if( uri.length() < REQUEST_SCHEMA.length( ))
+        return false;
+    return uri.compare( uri.length() - REQUEST_SCHEMA.length(),
+                        REQUEST_SCHEMA.length(), REQUEST_SCHEMA ) == 0;
```
hernando: second parameter can be std::npos

zeroeq/http/server.h
```diff
@@ -167,13 +187,22 @@ class Server : public zeroeq::Receiver
...
+    ZEROEQ_API bool handleGET( const std::string& event, const GETFunc& func );
+
+    /**
+     * @overload
+     * @param event the event name to serve during receive()
+     * @param schema describes data layout of event
+     * @param func the callback function for serving the GET request
+     */
+    ZEROEQ_API bool handleGET( const std::string& event,
+                               const std::string& schema, const GETFunc& func );
```
hernando: For empty events it would be more handy not having to pass an empty schema parameter.
Does it make sense to move scheme to the end of the parameter list with "" as a default value, or do you want to enforce an explicit schema for some reason?

zeroeq/http/server.h
```diff
@@ -48,8 +48,8 @@ class Server : public zeroeq::Receiver
     ZEROEQ_API Server( const URI& uri, Receiver& shared );
     ZEROEQ_API explicit Server( const URI& uri );
     ZEROEQ_API explicit Server( Receiver& shared );
-    ZEROEQ_API explicit Server();
-    ZEROEQ_API explicit Server( Server& shared )
+    ZEROEQ_API Server();
+    explicit Server( Server& shared )
```
hernando: ZEROEQ_API removed on purpose? If this function is internal tag it.

zeroeq/http/server.h
hernando: Forget about it

zeroeq/http/server.cpp
```diff
@@ -68,58 +98,71 @@ class Server::Impl : public detail::Sender
...
-        event = _toLower( event );
-        if( _subscriptions.count( event ) != 0 )
+        event = _toLower( _replaceAll( event, "/", "::" ));
+        if( _put.count( event ) != 0 )
             return false;
 
-        _subscriptions[ event ] = func;
+        _put[ event ] = func;
+        if( !schema.empty( ))
+            _schemas[ event ] = schema;
```
hernando: Throw if there's an schema and it's not the same (maybe a good reason to use jsoncpp, so you can compare schemas without failing due to whitespace)

zeroeq/http/server.h
tribal-tec: for inline functions we typically have no FOO_API declaration.

zeroeq/http/server.cpp
```diff
@@ -26,6 +32,30 @@ std::string _toLower( std::string value )
...
+    while( (pos = subject.find( search, pos )) != std::string::npos )
+    {
+         subject.replace( pos, search.length(), replace );
+         pos += replace.length();
+    }
+    return subject;
+}
+
+const std::string REQUEST_REGISTRY = "registry";
+const std::string REQUEST_SCHEMA = "schema";
```
tribal-tec: The slash is stripped by _getTypeName(), so yes, no slash here.

zeroeq/http/server.cpp
tribal-tec: ?

zeroeq/http/server.cpp
tribal-tec: You assume then that the schema is always JSON-formatted.

hernando: At least you should do a string comparison then. What does it mean to register GET and PUT with different schemas otherwise?

zeroeq/http/server.cpp
hernando: uri.compare( uri.length() - REQUEST_SCHEMA.length(),
                      std::npos, REQUEST_SCHEMA )

zeroeq/http/server.h
hernando: I always forget about this.

zeroeq/http/server.cpp
```diff
@@ -109,13 +115,18 @@ class Server::Impl : public detail::Sender
...
             return false;
 
         _put[ event ] = func;
-        if( !schema.empty( ))
+        {
+            const std::string& exist = _returnSchema( event );
+            if( !exist.empty() && schema != exist )
+                ZEROEQTHROW( std::runtime_error(
+                             "Schema registered for event differs: " + event ));
             _schemas[ event ] = schema;
```
hernando: Do no reassign if already the same.

tribal-tec:
Good now?

rdumusc:
+1 for me despite my question about the removePUT/GETs has been ignored. I feel that we should have waited until (if?) we had a concrete need for so many methods, but as you wish :-)

tribal-tec:
The remove... methods are only for symmetry reasons I guess. None of our code removes events from the server, so not even a remove() is needed. I can remove (haha) them actually.

PR #184, opened 2016-11-10
Use cppnetlib to implement HTTP server

tribal-tec:
The previous implementation has no support for multiple clients as requests
from different clients are de-multiplexed onto one stream socket.

cppnetlib implements this feature and maybe others which we don't handle now
as well. From past experience with RESTBridge, this seems the best approach
for now.


tests/http/server.cpp
```diff
@@ -9,34 +9,79 @@
...
 {
+namespace http = boost::network::http;
+// default client from cppnetlib is asynchronous, not handy for tests
+using HTTPClient =
+    http::basic_client< http::tags::http_default_8bit_tcp_resolve, 1, 1 >;
+
+using ServerReponse = http::basic_response<http::tags::http_server>;
+HTTPClient::response _buildResponse( const std::string& body = std::string( ))
+{
+    HTTPClient::response response;
```
favreau: It seems that you have Cors Headers implemented several times with the same values (_buildResponse, _buildError and _addCorsHeaders). Any chance this could be put in a single place?

zeroeq/http/httpHandler.h
```diff
@@ -0,0 +1,151 @@
...
+#include <string>
+#include <zmq.h>
+
+namespace zeroeq
+{
+namespace http
+{
+
+class HTTPHandler;
+typedef boost::network::http::server< HTTPHandler > HTTPServer;
```
rdumusc: cppnetlib server has the same HTTP* prefix as the locally defined classes - removing the HTTP prefix from local classes, or renaming this typedef, could help avoid any confusion

zeroeq/http/httpHandler.h
```diff
@@ -0,0 +1,151 @@
...
+        HTTPServer::response_header contentType;
+        contentType.name = "Content-Type";
+        contentType.value = "application/json";
+
+        response.headers.push_back( contentLength );
+        response.headers.push_back( contentType );
+        _addCorsHeaders( response );
+    }
+
+    void log( const HTTPServer::string_type& )
```
rdumusc: OK, could you add a small comment in the code saying so for future readers?

PR #188, opened 2016-12-09
Support custom HTTP URL endpoint for handled Serializables, expose hyphenated URLs for CamelCase event names

tribal-tec:
Use case:

1. Proper support for registering the "stream-to" endpoint for Deflect streaming in brayns by reusing the same serializable from Lexis, but with a different name that in the end shows up in the python notebooks.
2. The hyphenated URL is used as the method/object name in python instead of the generated typename from the schema.

I wasn't sure of creating (another!) overload for handle/handlePUT/handleGET with the customizable event name or what I did now with the default parameter to not break the API. Let me know what you prefer.

### DIFF
--- a/zeroeq/http/server.cpp
+++ b/zeroeq/http/server.cpp
@@ -22,10 +22,21 @@
 
 namespace
 {
-std::string _toLower( std::string value )
+// Transform camelCase to hyphenated-notation, e.g.
+// lexis/render/LookOut -> lexis/render/look-out
+// Inspiration: https://gist.github.com/rodamber/2558e25d4d8f6b9f2ffdf7bd49471340
+std::string _camelCaseToHyphenated( std::string camelCase )
 {
-    std::transform( value.begin(), value.end(), value.begin(), ::tolower );
-    return value;
+    std::string str( 1, tolower(camelCase[0]) );
+    for( auto it = camelCase.begin() + 1; it != camelCase.end(); ++it )
+    {
+        if( isupper(*it) && *(it-1) != '-' && islower(*(it-1)))
+            str += "-";
+        str += *it;
+    }
+
+    std::transform( str.begin(), str.end(), str.begin(), ::tolower );
+    return str;
 }
 
 // http://stackoverflow.com/questions/5343190
@@ -42,9 +53,9 @@ std::string _replaceAll( std::string subject, const std::string& search,
 }
 
 // convert name to lowercase with '/' separators instead of '::'
-void _convertEventName( std::string& event )
+void _convertEndpointName( std::string& endpoint )
 {
-    event = _toLower( _replaceAll( event, "::", "/" ));
+    endpoint = _camelCaseToHyphenated( _replaceAll( endpoint, "::", "/" ));
 }
 
 const std::string REQUEST_REGISTRY = "registry";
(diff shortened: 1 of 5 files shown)


zeroeq/http/server.h
```diff
@@ -93,9 +93,17 @@ class Server : public zeroeq::Receiver
...
     /** @name Object registration for PUT and GET requests */
     //@{
-    /** Handle PUT and GET for the given object. */
-    bool handle( servus::Serializable& object )
-        { return handlePUT( object ) && handleGET( object );}
+    /**
+     * Handle PUT and GET for the given object.
+     *
+     * @param object the object to update and serve on receive()
+     * @param event uses this as the URL endpoint instead of the default
```
rdumusc: should be 'use', not 'uses'. I also don't find 'event' very clear, but I don't know have a good suggestion. maybe 'address' or something like 'httpPath'?

zeroeq/http/server.cpp
```diff
@@ -129,12 +140,13 @@ class Server::Impl : public detail::Sender
         return foundPUT || foundGET;
     }
 
-    bool handlePUT( servus::Serializable& serializable )
+    bool handlePUT( servus::Serializable& serializable,
+                    const std::string& event )
```
eile: s/event/endpoint/

zeroeq/http/server.cpp
```diff
@@ -160,11 +172,12 @@ class Server::Impl : public detail::Sender
...
     {
         const auto func = [&serializable] { return serializable.toJSON(); };
-        return handleGET( serializable.getTypeName(), serializable.getSchema(),
-                          func );
+        return handleGET( event.empty() ? serializable.getTypeName() : event,
+                          serializable.getSchema(), func );
     }
 
     bool handleGET( std::string event, const std::string& schema,
```
eile: endpoint

zeroeq/http/server.h
```diff
@@ -114,9 +122,12 @@ class Server : public zeroeq::Receiver
      * The subscribed object instance has to be valid until removePUT().
      *
      * @param object the object to update on receive()
+     * @param event uses this as the URL endpoint instead of the default
+     *              servus::Serializable::getTypeName()
      * @return true if subscription was successful, false otherwise
      */
-    ZEROEQHTTP_API bool handlePUT( servus::Serializable& object );
+    ZEROEQHTTP_API bool handlePUT( servus::Serializable& object,
+                                   const std::string& event = std::string( ));
```
eile: overloads with (endpoint, object)

tribal-tec:
Updated

zeroeq/http/server.h
```diff
@@ -176,37 +199,46 @@ class Server : public zeroeq::Receiver
...
 
     /**
-     * Subscribe an event to serve HTTP GET requests.
+     * @overload
+     * @param object the object to serve during receive()
+     * @param endpoint use this as the URL endpoint instead of the default
+     *                 servus::Serializable::getTypeName()
+     */
+    ZEROEQHTTP_API bool handleGET( const std::string& endpoint,
+                                   servus::Serializable& object );
```
eile: const ..& object?

tribal-tec: Done

PR #189, opened 2016-12-15
Fix missing/wrong HTTP server socket notifications

tribal-tec:


zeroeq/receiver.cpp
```diff
@@ -96,51 +96,71 @@ class Receiver
...
-        case 0: // timeout; no events signaled during poll
-            return false;
+            switch( zmq_poll( sockets.data(), int( sockets.size( )),
+                              timeout == TIMEOUT_INDEFINITE ? -1 : timeout ))
+            {
+            case -1: // error
+                ZEROEQTHROW( std::runtime_error( std::string( "Poll error: " ) +
+                                                 zmq_strerror( zmq_errno( ))));
+            case 0: // timeout; no events signaled during poll
+                return haveData;
```
rdumusc: following the life of this variable was not easy but I think you got it quite right. However I still wonder if it can actually happen that the zmq_poll() gives a timeout even though we know that one of the socket still has data (i.e. haveData==true). And then, if this does occur, shouldn't we throw a runtime error instead or returning true? Otherwise we leave the socket in an undefined state like before this PR (i.e. we haven't read everything as the edge-triggered contract implies), correct?

CMakeLists.txt
```diff
@@ -9,7 +9,20 @@ list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMake
                               ${CMAKE_SOURCE_DIR}/CMake/common)
 include(GitExternal)
 
-set(ZEROEQ_DEB_DEPENDS libboost-test-dev libzmq3-dev)
+set(ZEROEQ_DEB_DEPENDS
+    libboost-test-dev
+    libzmq3-dev
+    # cppnetlib dependencies
```
rdumusc: Thanks for the addition! Not 2 spaces for indentation?

zeroeq/receiver.cpp
tribal-tec: zmq_poll() only returns 0 if there is no more data. As only the first call uses a timeout (any subsequent call has timeout == 0), it's guaranteed to not miss any data and we have to return true/false if we had data or not. Not sure that clears it up or if I got you right.

zeroeq/receiver.cpp
```diff
@@ -96,51 +96,71 @@ class Receiver
...
-        switch( zmq_poll( sockets.data(), int( sockets.size( )),
-                          timeout == TIMEOUT_INDEFINITE ? -1 : timeout ))
-        {
-        case -1: // error
-            ZEROEQTHROW( std::runtime_error( std::string( "Poll error: " ) +
-                                             zmq_strerror( zmq_errno( ))));
-        case 0: // timeout; no events signaled during poll
-            return false;
+            switch( zmq_poll( sockets.data(), int( sockets.size( )),
+                              timeout == TIMEOUT_INDEFINITE ? -1 : timeout ))
```
tribal-tec: Good point. Done

CMakeLists.txt
tribal-tec: Done

zeroeq/receiver.cpp
rdumusc: Not entirely sure too. What I was trying to determine is if 'case 0' can ever happen after the first call. It seems to me that since we only do another zmq_poll if(socket.revents & ZMQ_POLLIN), then the poll can't possibly return 0 in that case and that if it does, then that is a unexpected beahviour, not a regular return true.

PR #200, opened 2017-02-10
Feature spec: HTTP Server extension for generic REST interfaces

rdumusc:
I've tried to gather the requirements that we identified during our work on Tide html interface and propose an implementation. See formatted version [here](https://github.com/rdumusc/ZeroEQ/blob/7dd876345daca10e4c93fc5aa4de4aaf2285a433/doc/feature/httpServerREST.md).

For those interested, some more thoughts about ZeroEQ / REST in [this internal document](https://bbpteam.epfl.ch/project/spaces/display/~dumusc/ZeroEQ+REST). It is not include here as it does not fit the feature spec format.

### DIFF
--- a/doc/feature/httpServerREST.md
+++ b/doc/feature/httpServerREST.md
@@ -18,10 +18,12 @@ it usable as a general-purpose REST interface.
 
 None, cppnetlib 0.13 already offers the required functionality.
 
+# Proposition #1
+
 ## API
 
-To fullfill the abovementioned requirements with minimal API changes, we propose
-a generic interface as follows:
+To fullfill the requirements with minimal API changes, we propose a generic
+interface as follows:
 
     namespace zeroeq
     {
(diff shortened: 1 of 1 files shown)


rdumusc:
@ppodhajski FYI, can't add you to the list of reviewers (?)

eile:
@rdumusc : @ppodhajski was not part of this org. Invitation sent.

ppodhajski:
Stefam, I think you invited me again to BlueBrain instead of HBPVIS.


2017-02-12 14:59 GMT+01:00 Stefan Eilemann <notifications@github.com>:

eile:
Nope:
screen shot 2017-02-13 at 16 06 48

doc/feature/httpServerREST.md
```diff
@@ -0,0 +1,161 @@
...
+        std::string payload;
+        std::map<Header, std::string> headers;
+    };
+
+    /** HTTP REST callback with payload, returning a Response future. */
+    using RESTFunc = std::function<std::future<Response>(const std::string&)>;
+
+    /** HTTP REST callback with path + payload, returning a Response future. */
+    using RESTPathFunc = std::function<std::future<Response>(const std::string&,
+                                                             const std::string&)>;
```
hernando: As I understand it with this declarations you are leaving to the client the responsibility of implementing the asynchronous handling of requests.  If a client creates the future with std::async(std::deferred, ...) then the request is not handled asynchronously any more.
Is this deliberate for good reasons? Couldn't we hide the threading in the server internals?
If the server implementation is responsible of calling get in the future, I no good reason to require these function to return a future. The server could wrap in internally in the same way you've done in the first example.

doc/feature/httpServerREST.md
```diff
@@ -0,0 +1,161 @@
...
+        Code code = Code::OK;
+        std::string contentType;
+        std::string payload;
+        std::map<Header, std::string> headers;
+    };
+
+    /** HTTP REST callback with payload, returning a Response future. */
+    using RESTFunc = std::function<std::future<Response>(const std::string&)>;
+
+    /** HTTP REST callback with path + payload, returning a Response future. */
```
hernando: Explain better the parameters, it's not clear until the example.

doc/feature/httpServerREST.md
```diff
@@ -0,0 +1,161 @@
...
+};
+
+## Issues
+
+### 1: How many threads should the cppnetlib server use?
+
+_Resolution: Open:_ The server thread pool consists of a single thread by
+default. Using more threads may improve performance by accepting multiple
+concurrent requests but would require changes in the zeroeq inproc signaling
+mechanism.
```
hernando: As I see it, the point of using futures is that the cppnetlib pool can remain with 1 thread, the asynchronicity being provided by additional threads (in the current form of this specification, created by the user), but maybe I'm missing a detail here.

doc/feature/httpServerREST.md
```diff
@@ -0,0 +1,161 @@
...
+* Single endpoint for collections (e.g. windows/, files/, neurons/, etc. )
+* Allow filtering operations on collections (windows/selected, files?limit=10)
+
+## Dependency Changes
+
+None, cppnetlib 0.13 already offers the required functionality.
+
+## API
+
+To fullfill the abovementioned requirements with minimal API changes, we propose
```
rdumusc: Haha you guessed right, I'll reword this then

doc/feature/httpServerREST.md
```diff
@@ -0,0 +1,161 @@
...
+a generic interface as follows:
+
+    namespace zeroeq
+    {
+    namespace http
+    {
+    enum class Verb { GET, POST, PUT, PATCH, DELETE };
```
rdumusc: True, but I must say that after using both for a while I still find that upper-case for constants/enum makes the code easier to read. That's just a detail though.

doc/feature/httpServerREST.md
rdumusc: yes, will do

doc/feature/httpServerREST.md
rdumusc: Correct, the first idea was to add the possibility for the user to delay the response to a request without touching too much the server implementation. But what you suggest is also valid: the callback function(s) could be called directly from the cppnetlib thread(s), since it/they block on response.get() anyway. That might even make the API simpler.

eile:
I vote for option 1:
* It is simpler
* It has a clear separation of responsibities (App threading is managed by app if needed, allows for any model the app wants)
* Multithreading cppnetlib can still be done (with one single queue cppnetlib-main)

hernando:
Option 1 is deferring the problem for later and makes the code more verbose for the client in any situation, even if you don't care about asynchronous requests, which I'm not sure is necessarily better. But I wont' oppose that resolution.

eile:
Which problem is deferred to later?

eile:
For synchronous request we can have syntactic sugar to wrap the function result in a ready future, which we'll have for the legacy API anyhow.

hernando:
It defers the threading, despite cppnetlib already has it's own solution and the implementation will need to change anyway, but not a big deal.

I thought that has come to my mind is that something that needs to be verified for an efficient implementation to work with a single cppnetlib thread is whether it's possible to wait on multiple futures and file descriptors and the same time. This may be easier to implement if the implementation doesn't have to deal with std::futures but can use whatever synchronization mechanisms cppnetlib provides, which is only possible with option 2.

rdumusc:
So based on offline discussion, I am merging this proposition as it is. Unless there are more comments, we will use the first proposition as a base for doing the implementation.

PR #202, opened 2017-03-06
Switch to official cppnetlib

eile:


rdumusc:
Yeah! Good news!

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "cppnetlib"]
 	path = cppnetlib
-	url = https://github.com/BlueBrain/cpp-netlib
+	url = https://github.com/cpp-netlib/cpp-netlib
 	branch = 0.13-release
```
tribal-tec: master branch?

tribal-tec:
missing rebase

.gitmodules
eile: Sure?

rdumusc: I'm not sure, branch 0.13-release has your latest fixes for ICC that master does not have:
https://github.com/cpp-netlib/cpp-netlib/commits/0.13-release
https://github.com/cpp-netlib/cpp-netlib/commits/master
actually why not use an exact sha1? Using a development branch might break the build at any time.

PR #203, opened 2017-03-10
Implementation: HTTP Server extension for generic REST interfaces 

ppodhajski:


doc/Changelog.md
```diff
@@ -4,13 +4,13 @@
 
 * [203](https://github.com/HBPVIS/ZeroEQ/pull/203):
   HTTP server extension for generic REST interfaces.
-  A client can now:
+  Users can now:
     * use all standard HTTP verbs: GET, POST, PUT, PATCH, DELETE.
     * set a response header, including content type.
     * set a custom response payload for all action verbs.
-    * return a future and and avoid blocking of the main thread.
+    * return an std::future to avoid blocking the main thread.
```
eile: * use a std::future

eile: scratch that.

doc/Changelog.md
```diff
@@ -4,13 +4,13 @@
...
     * register a single endpoint for collections.
-    * filter collections using a query string with a GET request
+    * filter collections using a query string with GET requests.
```
eile: Style: no '.' at the end (most items are like that)

zeroeq/http/helpers.h
```diff
@@ -0,0 +1,27 @@
...
+{
+
+/** @return ready future wrapping the value passed. */
+template<typename T>
+std::future<T> make_ready_future( const T value )
+{
+    std::promise<T> promise;
+    promise.set_value( value );
+    return promise.get_future();
+}
```
eile: I feel somewhat strongly about not having this as part of the zeroeq API. It does not belong here, but to lunchbox/extra. I hid it on purpose in my CR for that reason, until we figure the above out.

rdumusc: Yes, it is missing from the standard. It could go to this future Extra project above ZeroEQ at some point, but now I see two solutions:
- make this header private like you had it and let users figure out how to make a ready Response.
- new alternative: keep the header public but change the function to:
std::future<Response> make_ready_response(...);
forwarding the arguments to the Response constructor, as that is the only use case for this helper. I think this would be quite nice to use!

eile: I'm fine with both.

kaabimg: You can avoid copy :
`
template <typename T>
std::future<T> make_ready_future(T&& value)
{
    std::promise<T> promise;
    promise.set_value(std::forward<T>(value));
    return promise.get_future();
}
`

zeroeq/http/response.h
```diff
@@ -0,0 +1,136 @@
...
+enum class Header
+{
+    LOCATION,
+    RETRY_AFTER,
+    LAST_MODIFIED,
+    CONTENT_TYPE
+};
+
+/** HTTP codes to be used in a Response. */
+enum Code
```
eile: enum class

zeroeq/http/helpers.h
rdumusc: @kaabimg you're perfectly right, this was an oversight. The new make_ready_response() function will correctly forward the arguments to the Response constructor.

zeroeq/http/response.h
rdumusc: I have a last-minute doubt about this. It felt somehow convenient (at least for the Tide unit tests) to have response.code directly comparable with integer values (200, 404) instead of having to use their enum name because many of those codes are well-known. What do you think?

rdumusc:
restest this please

PR #206, opened 2017-05-10
clang-format

eile:

PR #207, opened 2017-05-10
Cleanup type declarations to be symmetric with other function definitions.

eile:


eile:
6bdb3f9 is in #206. Will rebase once merged.

zeroeq/http/types.h
```diff
@@ -0,0 +1,38 @@
+
+/* Copyright (c) 2017, Human Brain Project
+ *                     Stefan.Eilemann@epfl.ch
+ */
+
+#pragma once
+
+#include <zeroeq/types.h>
```
rdumusc: not used?

eile:
Yes, we had discussed:
* the function typedefs for subscriber did not follow the same pattern
* I feel that one file, one type should be a rule with the exceptions of typedefs in types.h

Imo the only alternative would be making this sub-types of the class, which is done in other places.

eile:
Ping

zeroeq/http/types.h
eile: Is used, at least for some basic types like uint128_t

PR #208, opened 2017-05-12
Zeroconf announce for http::Server

eile:


zeroeq/http/server.cpp
```diff
@@ -176,6 +176,25 @@ class Server::Impl : public detail::Sender
...
+        const servus::Servus::Result& result =
+            _service.announce(uri.getPort(), getAddress());
+
+        if (!result)
+        {
+            ZEROEQTHROW(std::runtime_error("Zeroconf announce failed: " +
+                                           result.getString()));
+        }
+
         if (uri_.getHost() != uri.getHost() || uri_.getPort() == 0)
```
tribal-tec: this should be printed independently of the servus stuff

zeroeq/http/server.cpp
```diff
@@ -78,6 +80,7 @@ std::string _convertEndpointName(const std::string& endpoint)
 const std::string JSON_TYPE = "application/json";
 const std::string REQUEST_REGISTRY = "registry";
 const std::string REQUEST_SCHEMA = "schema";
+const std::string HTTP_SERVER_SERVICE = "_http._tcp";
```
tribal-tec: _zeroeqhttp._tcp? otherwise it looks very generic.

tribal-tec: Also, add some documentation in the header of the class that this features exists.

eile: I on purpose used the generic service for HTTP servers.

tribal-tec: Ah ok. Then mention it for future reference :)

zeroeq/http/server.cpp
eile: It is, unless Zeroconf announcement fails in which case the ctor throws anyways.

zeroeq/http/server.cpp
eile: doc added

zeroeq/http/server.cpp
eile: Correction, it isn't. Fixing.

eile: done

tribal-tec:
+1 for me. @hernando also should a have look

zeroeq/http/server.cpp
```diff
@@ -179,6 +179,25 @@ class Server::Impl : public detail::Sender
...
+        if (!_session.empty())
+            _service.set(KEY_SESSION, _session);
+
+        const servus::Servus::Result& result =
+            _service.announce(uri.getPort(), getAddress());
+
+        if (!result)
+        {
+            ZEROEQTHROW(std::runtime_error("Zeroconf announce failed: " +
+                                           result.getString()));
```
hernando: Do you really want to throw if the announcement fails?

eile: Since this is C&P from the Publisher, I guess so?

eile:
> Explain a bit more the motivation in the commit message.

Done

PR #210, opened 2017-05-24
http Server: fix handling of CORS preflight requests

rdumusc:


zeroeq/http/requestHandler.h
```diff
@@ -20,15 +21,31 @@ namespace http
 class RequestHandler;
 typedef boost::network::http::server<RequestHandler> HTTPServer;
 
+// HTTP headers for CORS responses
+enum class CorsResponseHeader
+{
+    ACCESS_CONTROL_ALLOW_HEADERS,
```
tribal-tec: lower-case and not prefix as this is an enum class?

rdumusc: there is no prefix, the header is this long ("Access-Control-Allow-Headers"). You're probably right about lower-case, although personally I still find upper-case more readable... since enum class Header in response.h also use upper case that makes it consistent.. but I can change it, no problem ;-)

PR #212, opened 2017-05-24
Fix build with clang 3.5 (Apple LLVM 6.0)

rdumusc:
However the http-server tests fail on my mac, I haven't been able to figure out why:

libc++abi.dylib: terminating with uncaught exception of type std::__1::bad_function_call: std::exception
unknown location:0: fatal error: in "get_serializable": signal: SIGABRT (application abort requested)
../ZeroEQ/tests/http/server.cpp:458: last checkpoint
unknown location:0: fatal error: in "get_event": memory access violation at address: 0x09f14cb0: no mapping at fault address
../ZeroEQ/tests/http/server.cpp:467: last checkpoint: "get_event" entry.

### DIFF
--- a/tests/http/server.cpp
+++ b/tests/http/server.cpp
@@ -223,7 +223,8 @@ private:
     void _checkImpl(const zeroeq::http::Method method,
                     const std::string& request, const std::string& data,
                     const Response& expected, const int line,
-                    std::map<std::string, std::string> requestHeaders = {})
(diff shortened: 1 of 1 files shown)


tribal-tec:
which OSX version do you have? for me the mac build work fine on latest sierra.

rdumusc:
10.10 Yosemite, c++ = Apple LLVM 6.0 (clang-600.0.54), zmq 4.1.4

PR #213, opened 2017-06-06
Client/server (req-rep) specification

eile:


eile:
See latest diff for a variant of asynchronous requests. The zmq API (req-rep-dealer-router) and doc is "flexible" for this type of use cases. I hope I got it now for the impl.

eile:
> There is an apparent contradiction between "Certain operations are by necessity synchronous" and "Requirements: Asynchronous remote procedure call".

The exchange is synchronous, the API is not. Will clarify.

eile:
> If the API is asynchronous it must return a future

No, it can also take a functor which is called on completion. I picked this approach since it does not require the reply object (if any) to be passed into the get().

eile:
I'll pick your use cases into the Examples section.

eile:
On reading them, maybe not: 1, 2 and 3 are special cases of 4:
1) is 4 with a return value
2) is 4 with parameters and a return value
3) is 4 with parameters

eile:
Or, to phrase it differently: The get Serializable is both your function name (the type uuid) and parameters (may be empty), and the ReplyFunc is your return value. With this you can do any remote call.

doc/feature/ReqRep.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Requirements
+
+* Asynchronous remote procedure call from a Client to a Server
+
+## API
+
+    using ReplyFunc = std::function<void(const void*, size_t)>;
+
+    class Client : public Receiver
```
hernando: Is the inheritance from Receiver important from the user's point of view or is it just convenient for the implementation? If it's the latter replace it with implementetion-defined in the spec.

doc/feature/ReqRep.md
```diff
@@ -0,0 +1,112 @@
...
+Certain operations are by necessity synchronous, RPC-like exchanges. This
+specification describes this client-server API based on zmq req-rep (dealer/router) sockets.
+
+## Requirements
+
+* Asynchronous remote procedure call from a Client to a Server
+
+## API
+
+    using ReplyFunc = std::function<void(const void*, size_t)>;
```
hernando: Refresh my memory, do Zerobuf objects check the type ID during deserialization? Otherwise we are lacking type safety here, however I don't have a good suggestion to fix it (except add the type check in Zerobuf).
In any case, if the server doesn't use Zerobuf or some other typesafe serialization engine, we are in trouble.

doc/feature/ReqRep.md
```diff
@@ -0,0 +1,112 @@
...
+## Issues
+
+### 1: Shall the handle function return a future like the http::Server?
+
+_Resolution: No:_ The http::Server manages internal thread(s) for the protocol
+handling. The zmq rep socket does not, and a multithreaded server would still call handle() synchronously from each worker thread.
+
+### 2: Is the client a receiver?
+
+_Resolution: Yes:_ The get method only sends the request, the reply is received in receive() and correlated to the request.
```
hernando: If a client issues millions of requests and doesn't call receive at all, what does it happen on the server?

doc/feature/ReqRep.md
```diff
@@ -0,0 +1,112 @@
...
+* Asynchronous remote procedure call from a Client to a Server
+
+## API
+
+    using ReplyFunc = std::function<void(const void*, size_t)>;
+
+    class Client : public Receiver
+    {
+        // ctors same as Subscriber
+        void get(const Serializable& request, ReplyFunc& func);
```
hernando: How are server side errors propagated to the client?

doc/feature/ReqRep.md
```diff
@@ -0,0 +1,112 @@
...
+        MemoryUnitPtr getData(const LODNode& node)
+        {
+            MemoryUnitPtr block;
+            _client.get({node, _uri},
+                        [&]( const void* data, const size_t size ){
+                            block = new AllocMemoryUnit(size);
+                            memcpy data to block
+                        });
+
+            _client.receive( TIMEOUT );
```
hernando: To me it looks a lot more natural getting a future from _client.get and calling future.get to resolve it. Is there a way to achieve this without falling into thread-safety traps, or there's a fundamental reason for not doing it?

doc/feature/ReqRep.md
```diff
@@ -0,0 +1,112 @@
...
+            while( _server.receive()) /*nop*/ ;
+        }
+
+        zeroeq::Server _server;
+
+        servus::Serializable::Data _getInfo(const void* data, const size_t size)
+        {
+            ConstURIPtr uri = livre::URI::create(data, size);
+            if( openAndCacheDatasource( *uri ))
+                return dataSource.getVolumeInfo().toBinary();
```
hernando: The binary buffer returned will be moved or copied?

doc/feature/ReqRep.md
```diff
@@ -0,0 +1,112 @@
...
+
+_Resolution: Yes:_ The get method only sends the request, the reply is received in receive() and correlated to the request.
+
+### 3: Is the untyped return serializable an issue?
+
+_Resolution: Open_ The return value data is not typed, since it requires no
+routing to a function. This requires that the client and server use the same
+object and version for this, i.e., the hard versioning of ZeroBuf wrt member
+layout is lost. The ID is not part of the ZeroBuf, so would need separate
+transport and checking in server/client.
```
hernando: I see, this "answers"  my first comment.

tests/common.h
```diff
@@ -1,5 +1,5 @@
 
-/* Copyright (c) 2014-2016, Human Brain Project
+/* Copyright (c) 2014-2017, Human Brain Project
```
hernando: Good that you renamed it, I never liked the name if this file.

hernando:
Why not call get, request? That name doesn't imply there's a return value at all.

Using a future and a template function would allow to introduce some type safety at least in the API. Even if type-safety is not implemented initially, it could be implemented later. The future also provides a natural way of propagating exceptions from the server and allows proper handling in the client. Right now it's not clear at all how errors are handled.

doc/feature/ReqRep.md
eile: Yes, it is important: The user can call a single receive() on a set of publishers, http::servers, clients and servers.

doc/feature/ReqRep.md
eile: See below: func( nullptr, 0 ) is called.

doc/feature/ReqRep.md
eile: When does the receive happen in your case? The app still needs to call it, i.e., future.wait() would deadlock in a single-threaded application. The user can still wrap it in a future with a lambda function.

doc/feature/ReqRep.md
eile: Moved, Serializable::Data has a shared_ptr + size.

doc/feature/ReqRep.md
eile: Implementation-specific. zmq has high-water marks on pending requests, in this case the server reply will block eventually.

doc/feature/ReqRep.md
eile: The first comment from my POV does not make sense. Which one do you mean?

eile: BTW: Would be happy to get KISS ideas on how to resolve this.

eile:
> Why not call get, request? That name doesn't imply there's a return value at all.

Good idea. Will change.

eile:
> future and a template function would allow to introduce some type safety

Some, but not all. The bigger issue is that the reply object changes memory layout and we get a client-server mismatch. Since it would still be the same class, a future would not catch (ha!) this. The reply type id needs to be part of the protocol.

doc/feature/ReqRep.md
hernando: I didn't even think about it because that's an implementation detail. I'd argue that mixing request/reply (which are synchronous) with pub/sub receivers or even receivers from different servers doesn't look like a good idea. I don't know about the implementation details from zmq, but isn't the call to receive the reply blocking? Can you still poll multiple request sockets for the reply?

doc/feature/ReqRep.md
hernando: But can the server block in a single reply if the client doesn't process it? Or the behaviour in this case is a socket parameter?

doc/feature/ReqRep.md
hernando: My first comment is the first comment from my review, which even from your point of view is unique because it's at the very top.

hernando:
The future forces the client to tell the implementation which is the expected type, the implementation still needs some mechanism to gurantee that (which at the moment is missing). But since it also depends on changing the protocol, it's not so different from the current proposal in terms of guarantees anyway.

doc/feature/ReqRep.md
eile: Wrong thread, I guess. It is not an implementation detail. Client::request() is async, and Client::receive() is the sync point. In the sync point the app handles the reply from the server, which is symmetric with the subscription handling.

doc/feature/ReqRep.md
eile: It's a socket parameter, and afaik a single reply won't block.

doc/feature/ReqRep.md
eile: I guess it's the one at the very bottom!?:

> Using a future and a template...

See discussion there.

eile:
> The future forces the client to tell the implementation which is the expected type, the implementation still needs some mechanism to gurantee that (which at the moment is missing). But since it also depends on changing the protocol, it's not so different from the current proposal in terms of guarantees anyway.

The implementation should not care about the *type*, but about the *type id* which is a stronger guarantee and the only one used in other parts of the API. One gains exactly nothing with the template method, because even when using the concrete type it does not prevent sending another type from the server. Note that the sync API in the first iteration used ```get(Serializable& req, Serializable& rep).```

doc/feature/ReqRep.md
hernando: That's not entirely satisfactory. In a request that has an empty reply (item 3 in the list from Raphael) how can I distinguish a success from a failure?

doc/feature/ReqRep.md
hernando: I don't have a strong opinion here. I was just wondering if it could be more natural with the future as Raphael suggested. Right now it looks quite symmetric with the pub, sub, but error handling still looks clunky this way. Callback oriented APIs normally have hooks also for errors.

eile:
reply typing addressed in a1193a7

eile:
See https://github.com/eile/ZeroEQ/tree/reqrep for WIP

eile:
Ping: What's left to discuss here?

zeroeq/subscriber.h
```diff
@@ -70,25 +70,6 @@ class Subscriber : public Receiver
...
-     *   fully qualified
-     * - filters session \<username\> or ZEROEQ_SESSION from environment if
-     *   zeroeq::DEFAULT_SESSION
-     *
-     * @param uri publisher URI in the format [scheme://][*|host|IP|IF][:port]
-     * @param session session name used for filtering of discovered publishers
-     * @throw std::runtime_error if ZeroConf is not available or if session name
-     *                           is invalid
-     */
-    ZEROEQ_API Subscriber(const URI& uri, const std::string& session);
```
tribal-tec: Needs version bump. we should already go to the version 1.0.0 now I guess.

eile: done, to 0.9 for now. Also squashed commits in preparation of rebase-and-merge

eile:
I'll eventually need an approval...

PR #217, opened 2017-07-11
Fix HTTP server response to CORS requests

rdumusc:


tests/http/server.cpp
```diff
@@ -269,6 +272,17 @@ class Client : public HTTPClient
...
+        // Print headers before error to help debugging
+        if (responseHeaderSize != expectedHeaders.size())
+        {
+            std::cout << "--- Response headers ---" << std::endl;
+            for (const auto& header : responseHeaders)
+                std::cout << header.first << ": " << header.second << std::endl;
+            std::cout << "--- Expected headers ---" << std::endl;
+            for (const auto& header : expectedHeaders)
+                std::cout << header.first << ": " << header.second << std::endl;
+        }
```
favreau: Is this something that you want to keep? Or is it a left over from debugging sessions?

rdumusc: I want to keep it to improve the life of the next person who has to debug a failing unit test here. Previously all you could see was: "At l.912 3 != 2" which does not help much :-)

favreau: Indeed ;-D Thanks for the explanation, and for the fix.

PR #219, opened 2017-07-18
Request-reply (Client-Server) support

eile:


zeroeq/detail/context.h
```diff
@@ -0,0 +1,21 @@
...
+#include <memory>
+
+namespace zeroeq
+{
+namespace detail
+{
+using ContextPtr = std::shared_ptr<void>;
+
+/** @return the current ZeroMQ context. Users need to hold onto this context to
+ *          extend its lifetime to all sockets created from the context.
```
hernando: This documentation is not precise. The context is created if it doesn't exist.

This function is a bit strange for such a simple name as getContext.
Why storing a weak reference to the context internally? It seems as if you wanted to guarantee that all sockets will share the same context implicitly as long there's one object holding a context and remove the context once the last object using it is destroyed. If that's the intention it should be documented somewhere. I would rename it to getOrCreateSharedContext()

tests/reqRep.cpp
```diff
@@ -0,0 +1,238 @@
...
+{
+    test::Echo echo("The quick brown fox");
+
+    zeroeq::Server server(zeroeq::NULL_SESSION);
+    zeroeq::Client client(zeroeq::URI(server.getURI()));
+
+    std::thread thread([&] { runOnce(server, echo); });
+
+    bool handled = false;
+    client.get(echo.getTypeIdentifier(), [&](const void* data,
```
hernando: This is something that was not clearly specified in the feature request. Client get/set methods were mentioned, but only in a example, and the example doesn't match this test. What's the value of this get method over Client::request, because I don't see many differences.

tests/reqRep.cpp
```diff
@@ -0,0 +1,238 @@
...
+    {
+        BOOST_CHECK(client.publish(test::Echo(test::echoMessage)));
+
+        BOOST_CHECK(!noServer.receive(100));
+        if (server.receive(100))
+        {
+            received = true;
+            break;
+        }
+    }
```
hernando: I'm confused. Isn't this for request/reply tests? Why Client has a method called publish? Why servers has a subscribe method? This doesn't resemble the feature spec at all.

I think I see. You just copied the test from pubSub and "commented" this part. If the partial commits don't make sense it's a waste of time to look into them, but looking at the full pull request at once is really hard because of its size. What should I expect from the rest of the commits and what can I expect to be less time consuming? I was already feeling this PR was going to be hard to review and this commit hasn't improved that feeling.

zeroeq/client.h
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * If the client is in the same session as discovered servers, it
+ * automatically connects to those servers. Requests are executed round-robin on
+ * all connected servers.
+ *
+ * A connection to a non-existing server is valid. Requests will be executed
+ * once the servers are available.
+ *
+ * A receive on any instance of a shared group will work on all instance
+ * and call the registered handlers.
```
hernando: Rephrase this sentence.

zeroeq/client.h
```diff
@@ -0,0 +1,139 @@
...
+     * @param request the request identifier
+     * @param data the payload data of the request
+     * @param size the size of the payload data
+     * @param func the function to execute for the reply
+     */
+    ZEROEQ_API void request(const uint128_t& request, const void* data,
+                            size_t size, ReplyFunc& func);
+
+    /** @return the session name that is used for filtering. */
+    ZEROEQ_API const std::string& getSession() const;
```
hernando: Should  this method be in Receiver?

zeroeq/client.h
```diff
@@ -0,0 +1,139 @@
...
+     * all servers are overloaded or no server is connected.
+     *
+     * @param request the request identifier and payload
+     * @param func the function to execute for the reply
+     */
+    ZEROEQ_API void request(const servus::Serializable& request,
+                            ReplyFunc& func);
+
+    /**
+     * Request the execution of the given data on a connected Server.
```
hernando: "Request the execution of data", that's a funny sentence

zeroeq/client.h
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * The reply function will be executed during receive(). May block of when
+     * all servers are overloaded or no server is connected.
```
hernando: block of, remove of

zeroeq/client.h
```diff
@@ -0,0 +1,139 @@
...
+    ZEROEQ_API Client(const URIs& uri, Receiver& shared);
+
+    /** Destroy this client. */
+    ZEROEQ_API ~Client();
+
+    /**
+     * Request the execution of the given data on a connected Server.
+     *
+     * The reply function will be executed during receive(). May block of when
+     * all servers are overloaded or no server is connected.
```
hernando: Is it true that this function may block (and same for the other overload)? Isn't receive the function that blocks?
If this function can block, then it should accept a timeout parameter or at least there should be a way to interrupt it.
How does zeromq work with REQ/REP sockets anyway? To me it seems a bit odd to make requests to an inexistent resource. I'm also missing a way to know that the client is connected to any server without having to use application level messages for that. It's also not clear to me whether a user may want to have a client which is not connected to anything after the constructors return.

tests/reqRep.cpp
eile: The first can be called from the object without knowing it's type, and the second one is static and can be called without an instance (but needs the type)

tests/reqRep.cpp
```diff
@@ -0,0 +1,238 @@
...
+        BOOST_CHECK_EQUAL(got.getMessage(), "Jumped over the lazy dog");
+        handled = true;
+    });
+
+    BOOST_CHECK(!handled);
+    client.receive(TIMEOUT);
+    BOOST_CHECK(handled);
+}
+
+#if 0
```
eile: That's gone now.

tests/reqRep.cpp
```diff
@@ -0,0 +1,238 @@
...
+BOOST_AUTO_TEST_CASE(zeroconf)
+{
+    if (!servus::Servus::isAvailable() || getenv("TRAVIS"))
+        return;
+
+    zeroeq::Client client(test::buildUniqueSession());
+    zeroeq::Server noServer(client.getSession());
+    zeroeq::detail::Sender::getUUID() =
+        servus::make_UUID(); // different machine
+    zeroeq::Server server(client.getSession());
```
eile: that's gone. I started with a c&p from the pubsub tests

zeroeq/client.h
eile: done

zeroeq/client.h
eile: Yes, it's true since the request sending will block. Receive() will block on at least one reception of a reply.

eile: I would add a timeout on an as-needed basis. One needs to use ZMQ_DONTWAIT and spin, afaik. In reality I expect this to be a non-issue.

eile: One can connect to a zeroconf session which will dynamically update the servers as they are available. The Monitor can be used to get notifications.

zeroeq/client.h
eile: done

zeroeq/client.h
eile: Monitor is a session-less receiver.

eile:
Retest this please.

eile:
Retest this please.

zeroeq/receiver.cpp
```diff
@@ -180,6 +177,22 @@ Receiver::~Receiver()
     _impl->remove(this);
 }
 
+Receiver::Receiver(Receiver&& from)
+    : _impl(from._impl)
+{
+    from._impl.reset();
+}
+
+Receiver& Receiver::operator=(Receiver&& from)
```
eile: It can't, because otherwise the Impl needs to be defined in the declaration since this._impl might need to be destructed.

rdumusc: But you are talking about doing it in the header, right? What I'm saying is that it can be done in the cpp file instead, as shown in the link above. You still need to forward-declare it, but it saves writing the implementation.

eile: Ah yes, didn't think (so far) as you can do it in the definition (and didn't click the link yet). Will try.

eile:
Retest this please.

eile:
ping?

PR #223, opened 2017-07-26
There is no reason for a subscriber to not connect to multiple publishers, so let's

eile:


zeroeq/subscriber.cpp
```diff
@@ -45,24 +45,27 @@ class Subscriber::Impl
...
     {
-        if (uri.getScheme() == DEFAULT_SCHEMA &&
-            (uri.getHost().empty() || uri.getPort() == 0))
+        for (const URI& uri : uris)
         {
-            ZEROEQTHROW(std::runtime_error(
-                std::string("Non-fully qualified URI used for subscriber")));
-        }
+            if (uri.getScheme() == DEFAULT_SCHEMA &&
+                (uri.getHost().empty() || uri.getPort() == 0))
```
rdumusc: this could read more smoothly with a helper function:
```
if (!isFullyQualified(uri))
{ throw }
```
Or replacing the whole block:
`checkFullyQualified(uri);`

eile: Good point - I'll do it in #219 to avoid merge conflicts

PR #225, opened 2017-07-28
Fix #224: Handle exceptions in server handlers

### LINKED ISSUE #224: throw in handle functions not handled
in http::Server and the Server from #219, throws from the handlers should be handled and an error reported to the client.

eile:


eile:
Only the last commit is relevant. Will rebase once #219 is merged.

zeroeq/server.cpp
```diff
@@ -77,22 +77,27 @@ class Server::Impl : public detail::Sender
...
-            if (!_send(&reply.first, sizeof(reply.first),
-                       hasReplyData ? ZMQ_SNDMORE : 0))
+                if (_send(&reply.first, sizeof(reply.first),
+                          hasReplyData ? ZMQ_SNDMORE : 0) &&
+                    hasReplyData)
+                {
+                    _send(reply.second.ptr.get(), reply.second.size, 0);
+                }
+            }
+            catch (...) // handler had exception
```
tribal-tec: catch std::exception to extract and reply the what()  before (...) ?

eile: Then we'll need to define error codes (e.g. special reply IDs) for the Client. I would vote for doing this when we have a use case.

PR #232, opened 2017-08-25
Fix #230: Handle zeroconf disconnects

### LINKED ISSUE #230: Handle zeroconf disconnects
Currently we only connect() newly found Publishers and Servers, but never disconnect them. Less important for pub-sub, but deadly for req-rep.

eile:


zeroeq/detail/receiver.h
```diff
@@ -11,15 +11,18 @@
...
 
+#include <algorithm>
+
 namespace zeroeq
 {
 namespace detail
 {
 /** Manages and updates a set of connections with a zeroconf browser. */
-class Receiver
+class Receiver : public servus::Listener
```
rdumusc: I have the impression that this class has the wrong name. Is it not a ConnectionManager instead of a Receiver?

zeroeq/detail/receiver.h
```diff
@@ -129,6 +141,29 @@ class Receiver
...
+            return false;
+
+        auto socket = i->second;
+        if (zmq_disconnect(socket.get(), zmqURI.c_str()) == -1)
+        {
+            ZEROEQINFO << "Cannot disconnect from " << zmqURI << ": "
+                       << zmq_strerror(zmq_errno()) << std::endl;
+        }
+
+        auto j = std::find_if(_entries.begin(), _entries.end(),
```
tribal-tec: use remove_if directly?

tribal-tec:
Unit test?

PR #236, opened 2017-09-18
Avoid CMake error in optional cppnetlib

eile:


tests/reqRep.cpp
```diff
@@ -111,9 +113,11 @@ BOOST_AUTO_TEST_CASE(serializable)
     BOOST_CHECK(client.receive(TIMEOUT));
     BOOST_CHECK(handled);
 
+    #if (ZMQ_VERSION >= 40104)
```
tribal-tec: clangformat broken??

eile: fixed. Apparently.

PR #237, opened 2017-10-05
Fix missing handling of CORS requests on schema endpoints

tribal-tec:


zeroeq/http/server.cpp
```diff
@@ -416,16 +416,29 @@ class Server::Impl : public detail::Sender
...
+            const auto endpoint = path.substr(0, path.find_last_of('/'));
+            if (_schemas.find(endpoint) == _schemas.end() ||
+                message.accessControlRequestMethod != Method::GET)
+            {
+                message.response = make_ready_response(Code::NOT_SUPPORTED);
+                return;
+            }
+            allowedMethods = "GET";
+        }
+        else if (!_methods[int(message.accessControlRequestMethod)].count(path))
```
rdumusc: I think this condition should go first, so that a registered endpoint that ends with /schema is still handled correctly.

zeroeq/http/server.cpp
```diff
@@ -416,16 +416,29 @@ class Server::Impl : public detail::Sender
         // remove leading '/'
         const auto path = message.request.path.substr(1);
 
-        if (!_methods[int(message.accessControlRequestMethod)].count(path))
+        std::string allowedMethods;
+        if (_endsWithSchema(path))
+        {
+            const auto endpoint = path.substr(0, path.find_last_of('/'));
+            if (_schemas.find(endpoint) == _schemas.end() ||
+                message.accessControlRequestMethod != Method::GET)
```
rdumusc: The entire if/else if/else with two return in the middle is quite complicated... maybe extracting these some of it to a helper function could also help, something like: if (isCorsRequestForRegisteredSchema(message)) {...}

PR #239, opened 2017-10-22
Fix for gcc-7 warning in jsoncpp.cpp.

ptoharia:


rdumusc:
test this please

rdumusc:
Thanks for the patch, but I think it's incorrect because it changes the behavior of the code...
Since jsoncpp is an external dependency that's bundled in, the best way to fix it would be to either:
- try to grab a more recent copy of jsoncpp from https://github.com/open-source-parsers/jsoncpp that has hopefully been fixed for gcc7
- or disable the -Wimplicit-fallthrough using a CMake compile flag that only applies to this specific file

ptoharia:
Ok, thanks @rdumusc !

PR #244, opened 2018-03-12
Added static_cast to int and missing macros

gonkc:
Hello, I'm Gonzalo Bayo, from the group working back in URJC/UPM with Pablo Toharia.
Right now I would like to use ZeroEQ in a custom C++ project in Windows (VC14) and '/ZeroEQ/zeroeq/http/requestHandler.cpp' doesn't compile. 
I have got the error: C4267: '-=' : conversion from 'size_t' to 'int', possible loss of data.
This error can be fixed casting 'size_t' to 'int' in the same file.
Also, test projects are not linking successfully ( errors LNK2001, LNK2019 and LNK1120 ) due a missing macros in some functions of ZeroEQ and ZeroEQHTTP projects.
This error can be fixed using appropriated macros (ZEROEQ_API and ZEROEQHTTP_API) in each case.
Regards

### DIFF
--- a/zeroeq/http/requestHandler.cpp
+++ b/zeroeq/http/requestHandler.cpp
@@ -22,12 +22,15 @@ namespace http
 {
 namespace
 {
-int _getContentLength(const HTTPServer::request& request)
+size_t _getContentLength(const HTTPServer::request& request)
 {
     for (const auto& i : request.headers)
     {
         if (i.name == "Content-Length")
-            return std::stoi(i.value);
+        {
+            int size = std::stoi(i.value);
+            return (size > 0) ? size : 0;
+        }
     }
     return 0;
 }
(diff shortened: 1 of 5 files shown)


zeroeq/http/requestHandler.cpp
```diff
@@ -139,7 +139,7 @@ struct ConnectionHandler : std::enable_shared_from_this<ConnectionHandler>
         }
 
         _body.append(&range[0], size);
-        _size -= size;
+        _size -= static_cast<int>(size);
```
hernando: I would change _size to be of type size_t instead of casting. You may need to change _getContentLength to return size_t and add error checking in line 30 to return 0 if the conversion yields a negative number.

gonkc: Cast of _size removed.
The type of the variable _size changed to size_t.
Function _getContentLength now returns a size_t.
Function _getContentLength checks the returned value to avoid negative values.

rdumusc (approved):

